PR #41, opened 2022-04-12
Fix/fluentd: subsystem is hardcoded instead of coming from an environment variable 

amit-mazor:
### FIX: fluent-bit-http chart 
Currently the subsystem name is configured to be `kubernetes.container_name` and doesnt allow any changes. 
Changing the subsystem kubernetes field to `come from an env`, allowing it to be override.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,8 @@
+# Changelog
+
+## v0.0.2 / 2022-04-19
+
+### Fluentd-http 
+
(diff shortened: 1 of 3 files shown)


avivgold098 (commented):
Please make `CHANGELOG.md` like https://github.com/open-telemetry/opentelemetry-specification/blob/main/CHANGELOG.md

avivgold098 (approved):

talknopf (approved):

PR #43, opened 2022-04-19
Add codeowners file

amit-mazor:
Adding CODEOWNERS file


talknopf (changes requested):
@amit-mazor I should be here too.

talknopf (approved):

PR #47, opened 2022-04-24
Generating integrations manifests from github workflows for manual de…

amit-mazor:
As part of the GitHub Actions worfklows runs,
the desired `integration's manifests will be generated` using the CX-Integrations CLI,
and will be saved under 'manifests/<integration-name>-manifests.yaml'. 

* By adding the generation of the yamls inside the workflows, we can be sure we always have the most updated yamls that match the last release.

* The generated yamls are only for 'fluentd-http' and 'fluent-bit-http' since the CLI supports only these two. The Coralogix integrations are going to be deprecated, therefore it's not supported here.

### DIFF
--- a/.github/workflows/fluent-bit-http-chart.yml
+++ b/.github/workflows/fluent-bit-http-chart.yml
@@ -45,5 +45,5 @@ jobs:
           cd ${{ env.CXINTEGRATIONS_CLI_PATH }}
           ./cx-integrations generate ${{ env.CHART_NAME }} --outputdir manifests
           git add ./manifests/${{ env.CHART_NAME }}-manifests.yaml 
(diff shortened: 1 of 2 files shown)


avivgold098 (changes requested):
Publish the artifacts after the `generate` command

avivgold098 (changes requested):
@amit-mazor There is not an available plugin to upload files during the steps? I less like this workflow.
Don't forget there is protection on master which required the one approval, then we should look for something which is more github workflow native

amit-mazor:
@avivgold098 do we want to keep working on it ? or should I close it ?

avivgold098:
@amit-mazor close

PR #48, opened 2022-04-25
FIX: fluentd buffer configuration and fluentbit set retry no limit, c…

amit-mazor:
## Fixes and Updates:

* set `logLevel` on 'error' instead of 'info' in **all integrations**. 
* set default values in **fluentd-http** README `override file example`.
* set `Retry_Limit` in **Fluent-bit** to no limit in order for it to keep retrying and not lose any data 
* update **Fluentd** `overflow_action` to 'throw_exception' instead of 'block' 
* set **Fluentd** `flush_thread_count` to 4 for parallelism of the outputs

### DIFF
--- a/fluentd/http/README.md
+++ b/fluentd/http/README.md
@@ -24,7 +24,7 @@ fluentd:
   - name: SUB_SYSTEM_SYSTEMD
     value: kubelet.service
   - name: ENDPOINT
-    value: api.eu2.coralogix.com
+    value: <put_your_coralogix_endpoint_here>
   - name: "FLUENTD_CONF"
     value: "../../etc/fluent/fluent.conf"
   - name: LOG_LEVEL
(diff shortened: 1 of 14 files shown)


oded-dd (dismissed):

talknopf (approved):

fluentd/http/README.md
```diff
@@ -63,3 +75,58 @@ Afterwards go to Grafana press the `Create` tab, then press `import`, and paste
...
+                "integrations": ["myintegration"]
+        },
+        "notify_every": 60,
+        "description": "Fluentd buffer is full, destination capacity is insufficient for your traffic.",
+        "active_when": {
+                "timeframes": [{
+                        "days_of_week": [
+                                0,
+                                1,
+                                2,
```
avivgold098, line 121: Should move to docs under `fluentd`

avivgold098 (changes requested):
1. Please update the changelog 
2. In-depth docs should keep outside the main readme  
3. Endpoints should document either in each integration or in the root README

avivgold098 (changes requested):
Please upadte endpoints

avivgold098 (approved):

PR #60, opened 2022-05-10
Feature/add image for standalone docker

talknopf:
added an additional flow to support stand alone fluentd setup for system using stand alone un-managed docker images.
also added support for hiredis output plugin.
removed useage of kubernetes metadata plugin


fluentd/standalone-image/Dockerfile
```diff
@@ -0,0 +1,22 @@
+ARG IMAGE_VERSION=v1.14.0-debian-1.0
```
amit-mazor, line 1: just to make sure, the image we usually use, the daemonset image didn't work here? it couldnt run with it ?

talknopf: correct, that is why i took the non daemonset image

.github/workflows/fluentd-standalone-image.yml
```diff
@@ -0,0 +1,45 @@
+
+name: Fluentd-Image
```
amit-mazor, line 2: I suggest to change the name to fluentd-standalone so we wont get confused since the daemonset image workflow has the same name

amit-mazor (approved):

PR #65, opened 2022-05-18
fixed wrong tag in config segment of fluentd http values.yml

talknopf:


juan-coralogix (dismissed):

amit-mazor:
@talknopf Please add a description , whats the bug and how do we fix it here

amit-mazor:
@talknopf Please add a description , whats the bug and how do we fix it here

fluentd/http/values.yaml
```diff
@@ -3,7 +3,7 @@ fluentd:
 
   image:
     repository: coralogixrepo/coralogix-fluentd-multiarch
-    tag: v0.0.5
+    tag: v0.0.6
```
amit-mazor, line 6: the new image [0.0.6] includes only a prometheus plugin  in a newer version, but it seems like Fluentd doesnt let us override the previous version, meaning this tag doesn't actually change anything currently so no need to update it here. 

If you want to change it, then we need to update he CHANGELOG

amit-mazor, line 6: the new image [0.0.6] includes only a prometheus plugin  in a newer version, but it seems like Fluentd doesnt let us override the previous version, meaning this tag doesn't actually change anything currently so no need to update it here. 

If you want to change it, then we need to update he CHANGELOG

fluentd/http/values.yaml
```diff
@@ -174,7 +174,6 @@ fluentd:
           applicationName ${record.dig("kubernetes", "#{ENV['APP_NAME']}")}
           subsystemName ${record.dig("kubernetes", "#{ENV['SUB_SYSTEM']}")}
           computerName ${record.dig("kubernetes", "host")}
-          timestamp ${time.strftime('%s%L')}
```
amit-mazor, line 177: why do we need to delete it ?

amit-mazor, line 177: why do we need to delete it ?

talknopf:
> @talknopf Please add a description , whats the bug and how do we fix it here

we are marking the logs with "container" tag but the kebernetes plugin tage is filtering with *.kubernetes.** tag.
so i fixed the filter

fluentd/http/values.yaml
```diff
@@ -164,7 +164,7 @@ fluentd:
         @type null
       </match>
 
-      <filter kubernetes.**>
```
talknopf, line 167: @amit-mazor this is where we fixed the bug

fluentd/http/values.yaml
amit-mazor: we were always using the timestamp field , its was always a part of the records being sent no ? Im not sure I understand

amit-mazor:
@talknopf I think we should still use the older image tag, 0.0.5 and not 0.0.6 for now, its not giving us anything new currently

talknopf:
@amit-mazor @avivgold098 @oded-dd 
Can we continue this one?

avivgold098 (approved):

amit-mazor (approved):

PR #73, opened 2022-08-16
Add Open Telemetry Collector as an agent

amit-mazor:


opel-collector/values.yaml
```diff
@@ -0,0 +1,95 @@
...
+    
+  config:
+    exporters:
+      coralogix:
+        endpoint: "${CORALOGIX_ENDPOINT}:9443"
+        private_key: "${CORALOGIX_PRIVATE_KEY}"
+        application_name: "${APP_NAME}"
+        subsystem_name: "${SUBSYSTEM}"
+    receivers:
+      prometheus: null
```
oded-dd, line 36: You should add the OPEL recievers

```
      otlp:
        protocols:
          grpc:
            endpoint: 0.0.0.0:4317
          http:
            endpoint: 0.0.0.0:4318
```

opel-collector/values.yaml
```diff
@@ -0,0 +1,95 @@
...
+    processors:
```
oded-dd, line 37:
```
      batch: {}
      k8sattributes:
        passthrough: false
        extract:
          labels:
          - key_regex: '(.*)'
      memory_limiter: null
```

opel-collector/values.yaml
```diff
@@ -0,0 +1,95 @@
...
+  - name: CORALOGIX_ENDPOINT
+    value: "tracing-ingress.eu2.coralogix.com"
+    
+  config:
+    exporters:
+      coralogix:
+        endpoint: "${CORALOGIX_ENDPOINT}:9443"
+        private_key: "${CORALOGIX_PRIVATE_KEY}"
+        application_name: "${APP_NAME}"
+        subsystem_name: "${SUBSYSTEM}"
```
oded-dd, line 34: APP_NAME_RESOURCE_KEY="k8s.namespace.name"
SUBSYTEM_NAME_RESOURCE_KEY="k8s.deployment.name:"

opel-collector/values.yaml
```diff
@@ -0,0 +1,95 @@
...
+        filter:
+          node_from_env_var: KUBE_NODE_NAME # The agent only discovers pods from the same host that it is running on
+        extract:
+          metadata:
+            - k8s.pod.name
+            - k8s.pod.uid
+            - k8s.deployment.name
+            - k8s.namespace.name
+            - k8s.node.name
+            - k8s.pod.start_time
```
oded-dd, line 50: Missing extensions
This is required for the collector liveness
```
    extensions:
      health_check: {}
      memory_ballast: {}
```

opel-collector/values.yaml
```diff
@@ -0,0 +1,95 @@
...
+    service:
```
oded-dd, line 51: service:
      extensions:
        - health_check
        - memory_ballast
      pipelines:
        traces:
          exporters:
          - coralogix
          processors:
          - k8sattributes
          - memory_limiter
          - batch
          receivers:
          - otlp

opel-collector/values.yaml
```diff
@@ -0,0 +1,95 @@
...
+            - k8sattributes
+            - memory_limiter
+            - batch
+          receivers:
+            - otlp
+            - zipkin
+            - jaeger
+
+  image: 
+    tag: 0.57.2
```
oded-dd, line 68: There is 0.58 and it also includes more k8sattributes

opel-collector/values.yaml
```diff
@@ -0,0 +1,95 @@
...
+
+  tolerations: 
+    - operator: Exists
+
+  ports:
+    metrics:
+      enabled: true
```
oded-dd, line 75: Relevant only in the case of prometheus

opel-collector/values.yaml
```diff
@@ -0,0 +1,95 @@
...
+      cpu: 100m
+      memory: 1G
+
+  serviceMonitor:
+    enabled: true
+
+  prometheusRule:
+    enabled: true
+    defaultRules:
+      enabled: true
```
oded-dd, line 88: Only relevant in case of prometheus

oded-dd (changes requested):

opel-collector/values.yaml
amit-mazor: I dont need, its by default, as long as I dont put 'null' or want to override the port, Its coming exactly like this by default

opel-collector/values.yaml
amit-mazor: why we dont want the filter and all of the metadata I used ?

opel-collector/values.yaml
amit-mazor: yes I wrote in the description that I will open a separated PR on upgrading the chart version and the app version

opel-collector/values.yaml
amit-mazor: it says it must be enabled in order to use the servicemonitor of the agent itself

opel-collector/values.yaml
amit-mazor: they are both enabled by default, 
the health check was added by default to the daemonset as the liveness and readiness probe with the endpoint `0.0.0.0:13133`, 
and the memory extension is configured on `400 mb` memory by default. Im not familiar enough with it in order to change it now so I left it with the default, i can change to to percentages instead of mb, any recommendation ?

opel-collector/values.yaml
amit-mazor: its exactly like I configured except that I also added the jaeger and zipkin recievers and disabled the logs and metrics pipelines

opel-collector/values.yaml
amit-mazor: its what I configured in the env vars:
appname=metadata.namespace
subsystem=metadata.name which will be the name of the deployment 
but I dont know why you put 'resource_key', in Coralogix docs it says just 'application_name'

opel-collector/values.yaml
```diff
@@ -0,0 +1,95 @@
...
+    valueFrom:
+      fieldRef:
+        fieldPath: metadata.name
+  - name: KUBE_NODE_NAME
+    valueFrom:
+      fieldRef:
+        apiVersion: v1
+        fieldPath: spec.nodeName
+  - name: CORALOGIX_ENDPOINT
+    value: "tracing-ingress.eu2.coralogix.com"
```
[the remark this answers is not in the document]
amit-mazor: I can change it but I think there should be a default endpoint that can be override, but if I take this env down, the whole deploy will fail since the exporter wont find the endpoint env

opel-collector/values.yaml
amit-mazor: oh ok I understand, its taking all of the labels instead only specific ones. 
And I still kept the `node_from_env_var` filter because its supposed to reduce the resources usage [by querying only the pods running with the agent on the same host] so I think we should use it

opel-agent/values.yaml
```diff
@@ -0,0 +1,85 @@
...
+  fullnameOverride: otel-agent
+
+  extraEnvs:
+  - name: CORALOGIX_PRIVATE_KEY
+    valueFrom:
+      secretKeyRef:
+        name: coralogix-otel-privatekey
+        key: PRIVATE_KEY
+  - name: APP_NAME
+    value: production # Can be any static value
```
amit-mazor, line 13: @oded-dd Apparently a dynamic value is not working as expected, its taking the namespace of the agent itself instead of the pods. 
Also I deleted the subsystem because its going to be deprecated and its taking the servicename by default

avivgold098 (dismissed):

opel-agent/README.md
```diff
@@ -0,0 +1,54 @@
...
+| EUROPE1 | `tracing-ingress.coralogix.com`          |
+| EUROPE2 | `tracing-ingress.eu2.coralogix.com`      |
+
+---
+**NOTE**
+
+The Open Telemetry Coralogix exporter requires the Coralogix private key. Therefore the following secret must be created: 
+
+* The `private key` appears under 'Data Flow' --> 'API Keys' in Coralogix UI.
+
```
oded-dd, line 35: Can you add a GIF image showing it?

opel-agent/Chart.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - Coralogix
+dependencies:
+  - name: opentelemetry-collector
+    version: "0.25.0"
+    repository: https://open-telemetry.github.io/opentelemetry-helm-charts
+sources:
+  - https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-collector
+maintainers:
+  - name: Coralogix
+    email: support@coralogix.com
```
oded-dd, line 18: Should change to patform@coralogix.com (or our group email)

opel-agent/values.yaml
oded-dd: I think it can be resolved with the `attributesprocessor`

```
  resource:
    attributes:
    - key: 
      action: 
      value: 
```

oded-dd (approved):

opel-agent/README.md
amit-mazor: added jpg, does it matter gif/jpg?

opel-agent/values.yaml
amit-mazor: I will try again, but at the moment it doesnt work, so lets close it for now, and I'll keep trying

avivgold098 (approved):

opel-agent/values.yaml
amit-mazor: I tried a lot, it didnt work, and I looked at the code itself and it seems not supported
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/coralogixexporter/client.go#L74

amit-mazor:
I will change to gif tomorrow , and also upgrade @oded-dd

oded-dd:
@amit-mazor upgrade should be done by adding the relevant config to `eng-observability`. I'd like also the work you did for eng-logging to move to `eng-observability` so we will not have different repos for logs/metrics/traces all should be under one repo `eng-observability`

PR #74, opened 2022-08-18
fluentbit version update

giltk:
1. change version file
2. change fluentbit base image (1.9.3 -> 1.9.7)
3. change systemd subsystemName to a dynamic value.
4. add net.keepalive off to eliminate some users errors


amit-mazor:
A few notes about upgrading the Fluent-Bit : 
1. If you upgrade the Fluentbit version , please go over the `release notes` and add to the PR description what was changed in these versions, are there any interesting new updates? did they solve any bug ? what has changed? 

2. If you update the image, please update the image tag also in the `values.yaml` file, and bump the version in the `Chart.yaml`, so a new chart release will be created.

fluent-bit/http/values.yaml
```diff
@@ -174,24 +174,26 @@ fluent-bit:
           Header                private_key ${PRIVATE_KEY}
           compress              gzip
           Retry_Limit           False
+          net.keepalive         off
```
amit-mazor, line 177: why do we want to add it?

fluent-bit/http/values.yaml
```diff
@@ -174,24 +174,26 @@ fluent-bit:
...
+      input-systemd.conf: |-
         [INPUT]
           Name systemd
           Tag host.*
           Systemd_Filter _SYSTEMD_UNIT=kubelet.service
-          Read_From_Tail On 
-          Mem_Buf_Limit 5MB  
+          Read_From_Tail On
+          Mem_Buf_Limit 5MB
+          DB /var/log/host-journal.db
```
amit-mazor, line 189: why do we need it ? not sure its really necessary for systemd logs

fluent-bit/http/values.yaml
giltk: some customers get the error : [error] [http_client] broken connection to api.coralogix.com:443 
because we don't support using the same connection. this option make sure that fluentbit wont use the same connection.

fluent-bit/http/values.yaml
giltk: i guess its not if we are using Read_From_Tail (read only new logs coming to journal). but if we remove that setting then even if the pod die, no systemd logs will be dropped.

amit-mazor (dismissed):

fluent-bit/http/values.yaml
giltk: reverted this change

amit-mazor (approved):

PR #77, opened 2022-08-23
Delete extra line in otel agent workflow

amit-mazor:


amit-mazor:
Closing this PR since it was opened for testing the workflow , 
found that the problem was the access key secret , it was updated for some reason a month ago, and probably got the wrong value. updated it now, and its working

PR #80, opened 2022-08-24
FIX: disabling metrics and logs pipeline, update READMEs

amit-mazor:
## FIXES and UPGRADES:
* **Changed the `pipeline order`** - now metrics and logs are null [disabled] in the pipeline `after the traces section`. By having them before the tracing, the are enabled automatically, so the order affects it.  
* Update `main readme` to include the new otel chart in addition to the fluent charts and add some small fixes
* Update `otel-agent readme`, to get some info from the main readme
* Update the agent `installation command` - after some checks, its impossible to use `--set` to override only specific envs, it requires to set the whole envs in the installation command which is very complicated. 
Therefore, it requires to create an `override.yaml` file which includes all of the envs in order to change an env.
* Update the otel-agent secret to be the same secret as the fluent charts secret, so there is no need now to have two secrets for logs and traces, with the same key inside. 
* Added `CHANGELOG` for the agent

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,26 +31,29 @@ For installation of each integration, please go inside each intergation's direct
 - [Fluentd-HTTP chart](https://github.com/coralogix/eng-integrations/blob/master/fluentd/http/README.md)
 - [Fluent-bit-Coralogix chart](https://github.com/coralogix/eng-integrations/blob/master/fluent-bit/coralogix/README.md)
 - [Fluent-bit-HTTP chart](https://github.com/coralogix/eng-integrations/blob/master/fluent-bit/http/README.md)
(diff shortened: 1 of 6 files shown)


avivgold098 (changes requested):
@amit-mazor make changelog, also please move the charts to folder call `charts` so it will be separated from the example. I think it will require also update the github action paths

amit-mazor:
@avivgold098 The issue with the changelog is that it hasn't been updated since a long time ago, there are 6 new versions here of Fluentbit for example, that are not documented in the changelog. 
So I think it should be updated before adding this new release

avivgold098:
> check

So lets have specific one for otel

amit-mazor:
@avivgold098 sounds good, added a new dedicated changelog, its under the opel-agent dir

avivgold098 (approved):

PR #82, opened 2022-08-29
Fix Context

juan-coralogix:
context for GitHub action is wrong


SirNatan1 (approved):
comperd to other workflows and seems it's needed the path to the docker file in order to build

PR #84, opened 2022-09-01
Upgrade fluentd k8s metadata filter

amit-mazor:
Fluentd is using kubeclient version 4.9.3 [also in fluentd version 1.15] , 
and they created a workaround to solve it in fluentd with version 2.11.1 of fluent-plugin-kubernetes_metadata_filter. 
so Im setting here the specific version. 

* a newer version of this plugin is automatically enabled in fluentd 1.15, but im not upgrading the whole version now, so this is a quick fix to solve this bug for now, and we can upgrade the whole version in a spearated PR. 

see: 
https://github.com/fluent/fluentd-kubernetes-daemonset/issues/1361
https://github.com/fluent/fluentd-kubernetes-daemonset/issues/1368


doryer:
@amit-mazor We aim to solve an issue using this upgrade. or just updating the plugin

amit-mazor:
@doryer Updating the plugin is a workaround they did to solve this issue for now, as you can see here https://github.com/fluent/fluentd-kubernetes-daemonset/issues/1368. 
they need to include the kubeclient version 4.10.0 in their fluentd images which they currently dont . 

so we need to update the plugin, I just meant that we can upgrade the whole fluentd version from 1.14 to 1.15 and will upgrade this plugin on the way , but instead of upgrading everything right now, im just updating the specific plugin. upgrading the whole image can be done separately

oded-dd (approved):

PR #87, opened 2022-09-06
Upgrade opentelemetry collector app and chart version

amit-mazor:
Chart updates : 
* Default resources decreased - cpu is now 256m instead of 1core, and memory is now 512Mi instead of 2Gi [we set the limits to ourselves so its ok] 
* they added support for running the collector as a statefulset

App version updates:
* They deprecated some configs that we didnt use , and added some new features, not supposed to affect us and to break something

### DIFF
--- a/opel-agent/CHANGELOG.md
+++ b/opel-agent/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## OpenTelemtry-Collector
 
+### v0.0.3 / 2022-09-06
+
+* [UPRADE] Upgrading chart version from 0.25.0 to 0.30.0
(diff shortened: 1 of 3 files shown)


oded-dd (commented):
@amit-mazor I'd like us to expose the image (as part of the values) and not use the docker one. Meaning we should mirror the latest image to our repo (public) and use it as part of the cart values

oded-dd (approved):

amit-mazor:
@oded-dd do you mean pull their image and push it to our dockerhub public repo [coralogixrepo] ? 

what is the goal of doing it ?

amit-mazor:
Merging this, will take care of the image separately

PR #88, opened 2022-09-07
Ecs fix

SirNatan1:
Fixed the config for fluentd in ec2 usecases added:
- Dynamic app/subsystem name capeability
- removed the record for tag
- sending ecs-agent and fluentd logs to stdout

Fixed the config for firelens:
- sending fluentd and "agent" to stdout
- added detect exceptions


juan-coralogix:
Tested in EC2 and Fargate with Java Exceptions and Dynamic app and subsystem. 

Logs where sent correctly.

The only "thing" is that on fargate multiline skips the <CR> but its good.

Fargate:
image
EC2:
image

ShauliSolomovich (approved):

PR #89, opened 2022-09-08
Push the opentelemtry image to Coralogix docker repo

amit-mazor:


oded-dd (approved):

amit-mazor:
Didnt merge it because when I tried to use the 59 version from own repo, in our daemonset in dev environment, we got 'exec format error' although all of the dockerhub 59 version includes the same archs as I include here. 
Once using 'latest' tag from our own repo, it works. 
so something is weird here and I dont want to merge it to prevent a situation where we use the image from our repo and it doesnt work. 

so closing it for now, since we are anyway going to upgrade, maybe its a bug in this version.

PR #90, opened 2022-09-08
Add support for metrics in the opentelemtry coralogix agent

amit-mazor:
* Updated environment variables to support the metrics endpoints
* Updated readme - how to disable metrics
* Updated the readme section on the podmonitor 
* Removed the receivers section since its configuring the prometheus receiver by default 
* Updated serviceMonitor to podMonitor since serviceMonitor is not supported in agent mode


oded-dd (approved):

opel-agent/README.md
```diff
@@ -38,18 +40,40 @@ helm upgrade otel-coralogix-agent coralogix-charts-virtual/opentelemetry-coralog
...
-| Region  | Traces Endpoint
-|---------|-----------------------------------------------|
-| USA1	  | `tracing-ingress.coralogix.us:9443`           |
-| APAC1   | `tracing-ingress.app.coralogix.in:9443`       |
-| APAC2   | `tracing-ingress.coralogixsg.com:9443`        |
-| EUROPE1 | `tracing-ingress.coralogix.com:9443`          |
-| EUROPE2 | `tracing-ingress.eu2.coralogix.com:9443`      |
+| Region  | Traces Endpoint                               | Metrics Endpoint	
+|---------|-----------------------------------------------|-----------------------------------------------
+| USA1	  | `tracing-ingress.coralogix.us:9443`           | `otel-metrics.coralogix.us:443`
```
avivgold098, line 72: Why do we need to mention the port? this is native SSL one

opel-agent/values.yaml
```diff
@@ -16,17 +16,19 @@ opentelemetry-collector:
...
     exporters:
       coralogix:
-        endpoint: "${CORALOGIX_ENDPOINT}"
+        endpoint: "${CORALOGIX_TRACES_ENDPOINT}"
         private_key: "${CORALOGIX_PRIVATE_KEY}"
         application_name: "${APP_NAME}"
-    receivers:
-      prometheus: null # If you want to enable the prometheusrules/servicemonitor, this line must be deleted, and also the line above ['receivers:']
+        metrics:
+          endpoint: "${CORALOGIX_METRICS_ENDPOINT}"
```
avivgold098, line 31: Why not just user giving the suffix? and we will control on the prefix wether is logs / metrics /traces

opel-agent/values.yaml
```diff
@@ -71,12 +80,12 @@ opentelemetry-collector:
...
       verbs: ["get", "watch", "list"]
 
-  # If using Prometheus, the following part can be enabled in order to monitor the agent:
+  # In order to enable podMonitor, following part must be enabled in order to expose the required port:
   # ports:
   #   metrics:
   #     enabled: true
 
-  # serviceMonitor:
+  # podMonitor:
```
avivgold098, line 88: What is the motivation?

avivgold098 (changes requested):

opel-agent/values.yaml
```diff
@@ -49,7 +51,14 @@ opentelemetry-collector:
...
             - jaeger
-        metrics: null
+        metrics:
+          exporters:
+            - coralogix
+          processors:
+            - memory_limiter
+            - batch
+          receivers:
+            - prometheus
```
avivgold098, line 61: We should include both otlp and prometheus

avivgold098 (commented):
No need to mention prometheus receiver config? or it coming from the upstream chart? 
I'm not sure it's clear how to use it

opel-agent/README.md
amit-mazor: its required according to our exporter docs : 
https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/coralogixexporter

if I delete it is throws `missing port in address`

opel-agent/values.yaml
amit-mazor: I think it might be confusing, they might confuse and put the whole endpoint instead of just part of it, or miss some of it, I think it should be all or nothing, maybe only the region itself like i did in firehose, but i think putting `eu2.coralogix.com:443` can be a bit confusing.

amit-mazor:
@avivgold098 the prometheus receiver config is by default yes , coming from the upstream chart

avivgold098:
> @avivgold098 the prometheus receiver config is by default yes , coming from the upstream chart

@amit-mazor So let's have a comment on it

opel-agent/values.yaml
avivgold098: Okay, can you try without the port?

opel-agent/README.md
avivgold098: @amit-mazor Thx

opel-agent/values.yaml
amit-mazor: here I commented only about the endpoint, not related to the port. 
about the port, as I commented above: if I delete it, it throws missing port in address

opel-agent/values.yaml
```diff
@@ -49,7 +51,15 @@ opentelemetry-collector:
...
             - jaeger
-        metrics: null
+        metrics:
+          exporters:
+            - coralogix
+          processors:
+            - memory_limiter
+            - batch
+          receivers:
+            - prometheus # The Prometheus receiver is configured by default
```
amit-mazor, line 61: @avivgold098 added this comment or, or do you think it should be in another place?

avivgold098 (approved):

PR #95, opened 2022-09-19
fix: changed the image tag from latest to ver

SirNatan1:
Changed the image tag from latest to specific version.


SirNatan1:
a mistake in the version of the multi-arch

PR #103, opened 2022-11-09
Fargate metadata

juan-coralogix:
Fargate integration now includes a RB script that pulls docker metadata from V4 Api  and appends it to each log.


fluentd/aws-ecs/image/fargate.rb
```diff
@@ -0,0 +1,48 @@
...
+      record['docker'] = {
+        'id' => metadata['DockerId'],
+        'name' => metadata['Name'],
+        'container_hostname' => metadata['DockerName'],
+        'image' => metadata['Image'],
+        'image_id' => metadata['ImageID'],
+        'labels' => metadata['Labels']
+      }
+      record.delete('container_name')
+      record.delete('container_id')
```
giltk, line 42: I think we should remove the two other fields that gets generated in the docker metadata:
- ecs_task_arn
- ecs_task_definition

giltk: also i think docker.name and docker.container_hostname in fargate will always be the same name so we don't need both.

fluentd/aws-ecs/image/firelens.conf
```diff
@@ -49,7 +56,7 @@
       @type memory
       compress gzip
       flush_thread_count 4
-      chunk_limit_size 6MB
+      chunk_limit_size 16MB
```
giltk, line 59: why the limit size change?

giltk (changes requested):

fluentd/aws-ecs/image/fargate.rb
juan-coralogix: task_definition I think is helpful due to the version. But what ever you think.

giltk: we also have the version in another docker label so no need for the ecs metadata one.

juan-coralogix: Removed the duplications.

giltk:
@juan-coralogix creates a new branch to have less clutter

PR #106, opened 2022-11-18
[otel-agent] add self monitoring

povilasv:


otel-agent/values.yaml
```diff
@@ -78,13 +76,34 @@ opentelemetry-collector:
...
+            endpoint: 0.0.0.0:14250
+          thrift_http:
+            endpoint: 0.0.0.0:14268
+          thrift_compact:
+            endpoint: 0.0.0.0:6831
+      prometheus:
+        config:
+          scrape_configs:
+            - job_name: opentelemetry-collector
+              scrape_interval: 10s
```
giltk, line 99: @povilasv i think a scrape of 10s is very fast, should be 30s or even 60s.

oded-dd (approved):

PR #108, opened 2022-11-22
fix: new repository design.

giltk:
- created three main folder: logs,metrics,traces.
- modified the helm charts workflow to only happen on specific paths.


giltk:
@povilasv you are right :) but still i think this separation would be best

giltk:
closing this PR and opening a new one for less clutter

PR #110, opened 2022-11-28
Upgrade chart and app version to v0.66.0

amit-mazor:
* Upgraded chart version - there were no changes in the chart itself, but it upgrades the otel collector version from 64 to 66. 
* Updated changelog both retro and for current version
* version v0.66.0 - it has a few small updates like: Add AWS CloudWatch Receiver, add snmp receiver, small bug fixes

### DIFF
--- a/otel-agent/CHANGELOG.md
+++ b/otel-agent/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## OpenTelemtry-Agent
 
+### v0.0.11 / 2022-11-28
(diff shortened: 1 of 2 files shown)


oded-dd (commented):
We need to fix the collector dashboard it is broken. Also please consult with Or on what's needed to be added to our opensource repositories as we need to comply to some rules

amit-mazor:
Added CLA, its applied only to new PRs thats why its not here, 
and the dashboard I'll open a separated PR

PR #115, opened 2022-11-29
Update otel agent dashboard json and add appname subsystem variables

amit-mazor:
Added variables for appname and subsystem and added them to all metrics that have these labels. (queue metrics for instance dont have these labels so no need)

### DIFF
--- a/otel-agent/dashboard/otel-agent-dashboard.json
+++ b/otel-agent/dashboard/otel-agent-dashboard.json
@@ -1169,27 +1169,25 @@
           "expr": "sum by (processor)(rate(otelcol_processor_batch_batch_send_size_sum[10m]))",
           "hide": false,
           "interval": "",
-          "l          }
-            ]
-          }
-        },
-        "overrides": []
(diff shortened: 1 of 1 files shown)


amit-mazor:
* About the queue metrics, I was confused between the prometheus and otel specification, fixed now, both in the variables and in the panels themselves. 
* Splitted the panels of the `queues` so we can see clearly the capacity vs current size in each node. 
* Added `$__rate_interval`
* Removed the batch panel for now until we figure out why the values show us small values compared to batch processor configuration

oded-dd (approved):

amit-mazor:
@povilasv I dont see exporter_failed_logs metrics, maybe because we still dont export logs so this metric doesnt exist , but anyway added the failed_metrics next to the spans, 
and changed the queue panels again a bit

amit-mazor:
Pushing since the current json in master doesnt work and is not updated at all, 
If we will want to change something , I'll open a new PR

PR #116, opened 2022-11-30
repo redesign only logs and metrics folders

giltk:


amit-mazor:
@giltk Please add a description to this PR describing the new design , new updates / changes

giltk:
@povilasv updated README links
@amit-mazor updated with commit message

.github/workflows/fluent-bit-coralogix-chart.yml
```diff
@@ -5,10 +5,11 @@ on:
   push:
     branches: master
     paths:
-    - 'fluent-bit/coralogix/**'
+    - 'logs/fluent-bit/http/Chart.yaml'
+    - 'logs/fluent-bit/http/values.yaml'
```
amit-mazor, line 9: this workflow is specific for fluentbit-coralogix and not http, so the paths need to be '/logs/fluent-bit/coralogix' , also I would keep it like the origin with **

.github/workflows/fluent-bit-http-chart.yml
```diff
@@ -5,10 +5,12 @@ on:
   push:
     branches: master
     paths:
-    - 'fluent-bit/http/**'
+    - 'logs/fluent-bit/http/templates/**'
+    - 'logs/fluent-bit/http/Chart.yaml'
+    - 'logs/fluent-bit/http/values.yaml'
```
amit-mazor, line 10: same here, I would keep the ** , less lines , more arranged

.github/workflows/fluentd-coralogix-chart.yml
```diff
@@ -5,10 +5,11 @@ on:
   push:
     branches: master
     paths:
-    - 'fluentd/coralogix/**'
+    - 'logs/fluentd/coralogix/Chart.yaml'
+    - 'logs/fluentd/coralogix/values.yaml'
```
amit-mazor, line 9: same here

.github/workflows/fluentd-http-chart.yml
```diff
@@ -5,10 +5,11 @@ on:
   push:
     branches: master
     paths:
-    - 'fluentd/http/**'
+    - 'logs/fluentd/http/Chart.yaml'
+    - 'logs/fluentd/http/values.yaml'
```
amit-mazor, line 9: and same here

.github/workflows/fluent-bit-coralogix-chart.yml
amit-mazor: we can add:
```
    paths-ignore: 
          - '**.md'
```
or something like this:
```
    paths:
      - 'logs/fluent-bit/http/**'
      - '!logs/fluent-bit/http/README.md**'
```
so it will ignore the readme, but everything else in the chart should trigger it, I dont see a reason why not : )

README.md
```diff
@@ -27,11 +29,9 @@ helm repo update
...
 
 For installation of each integration, please go inside each intergation's directory:
-- [Fluentd-Coralogix chart](https://github.com/coralogix/eng-integrations/blob/master/fluentd/coralogix/README.md)
-- [Fluentd-HTTP chart](https://github.com/coralogix/eng-integrations/blob/master/fluentd/http/README.md)
-- [Fluent-bit-Coralogix chart](https://github.com/coralogix/eng-integrations/blob/master/fluent-bit/coralogix/README.md)
-- [Fluent-bit-HTTP chart](https://github.com/coralogix/eng-integrations/blob/master/fluent-bit/http/README.md)
-- [OpenTelementry-Agent chart](https://github.com/coralogix/eng-integrations/blob/master/otel-agent/README.md)
+- [Fluentd-HTTP chart](https://github.com/coralogix/telemetry-shippers/blob/master/logs/fluentd/http/README.md)
+- [Fluent-bit-HTTP chart](https://github.com/coralogix/telemetry-shippers/blob/master/logs/fluent-bit/http/README.md)
+- [OpenTelementry-Agent chart](https://github.com/coralogix/telemetry-shippers/blob/repo-redesign/otel-agent/README.md)
```
amit-mazor, line 34: what about fluentbit-coralogix and the prometheus ecs ?

README.md
```diff
@@ -54,7 +54,7 @@ data:
   PRIVATE_KEY: <encrypted-private-key>
 kind: Secret
 metadata:
-  name: coralogix-keys
+  name: integrations-privatekey
```
amit-mazor, line 57: in logs and metrics its correct that the secret is called `integrations-privatekey`, but in otel integration its called 'coralogix-keys' so it can be confusing if we dont see the otel-agent readme. 
@povilasv I think we should change it in otel so it will match the secret name of the other integrations.

amit-mazor:
Added a few comments, but besides that, I think the new design is great! much more clear and organized

README.md
giltk: i removed them because i dont want people to use them, i will keep them and add a DEPRECATED on the READme

.github/workflows/fluent-bit-coralogix-chart.yml
giltk: took your advice, used paths-ignore

README.md
giltk: @amit-mazor already talked with @oded-dd , after this RP is merged I will change to coralogix-keys in the fluentd/bit integrations as well.

PR #117, opened 2022-11-30
[otel-agent] add zpages and increase timeout

povilasv:
Also increase timeout to 30s as some folks had issues pushing log data and default timeout (5s) is rather strict


amit-mazor:
Lets update the CHANGELOG so it will be synchronized with our versions.

amit-mazor (approved):

PR #119, opened 2022-12-06
Fix india coralogix endpoint

amit-mazor:
Current india endpoint is missing the app , updating , also we need to update in the upstream , https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/coralogixexporter


SirNatan1 (approved):
Thanks Amit!

PR #126, opened 2022-12-08
[otel-agent] fix hostmetrics scrape errors

povilasv:


oded-dd (approved):
I assume we plan to add it to the upstream :)

PR #133, opened 2022-12-21
Set node_from_env_var filter for k8sattributes

amit-mazor:
### Adding node_from_env_var filter for the k8sattributes processor

By adding this filter, the agent will query pods running on the same node only.
`It reduces resource usage`. 

As you can see in the following img , the amount of hosts the agent is querying before using this filter was much higher, 
which affected the resources usage. 

image

### DIFF
--- a/otel-agent/CHANGELOG.md
+++ b/otel-agent/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### v0.0.18 / 2022-12-21
 
-* [FIX] Set node_from_env_var filter for k8sattributes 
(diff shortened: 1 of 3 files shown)


otel-agent/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## OpenTelemtry-Agent
 
+### v0.0.18 / 2022-12-21
+
+* [FIX] Set node_from_env_var filter for k8sattributes
```
avivgold098, line 7: I will add, reducing memory usage for those using k8sattributes

avivgold098 (changes requested):
Looking fine, small comment

avivgold098 (approved):

PR #135, opened 2022-12-26
Addeed: latest tag to workflow and fixed ver file

SirNatan1:
fixed Version file and workflow


amit-mazor (approved):

amit-mazor:
@SirNatan1 please change the name of the file to `otel-ecs-image.yaml` so it will be clear its building an image and not a chart

PR #138, opened 2023-01-09
Update prometheus job_name and serviceMonitor

amit-mazor:
### Updates and Fixes:  
* When using both the `otel-agent` and the `infrastructure-collector` we need to separate the **prometheus jobs** in order to be able to monitor both and have separate alerts on both collectors. 
* PodMonitor cant be enabled in deployment mode so changing to `ServiceMonitor`
* Added `CHANGELOG`

### DIFF
--- a/otel-infrastructure-collector/CHANGELOG.md
+++ b/otel-infrastructure-collector/CHANGELOG.md
@@ -1,0 +1,8 @@
+# Changelog
+
+## OpenTelemtry-Infrastructure-Collector
+
+### v0.0.2 / 2023-01-09
+ 
+* [CHANGE] Change Prometheus job name for the collector itself
+* [FIX] Change PodMonitor to ServiceMonitor since PodMonitor is not enabled in Deployment mode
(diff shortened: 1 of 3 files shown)


nicolastakashi:
> @amit-mazor what is the reason for ServiceMoitor, we have to make sure they are using endpoint-slice API and not endpoints API. See issue - [prometheus-operator/prometheus-operator#3862](https://github.com/prometheus-operator/prometheus-operator/issues/3862)
> 
> On cluster which more than 1k nodes, k8s needed to use the endpoint slice

I guess it's more related to the number endpoints under and Endpoint object than the number of nodes, if I'm not wrong the endpoint controller is configured by default to ensure that and endpoint/endpoint slices doesn't have more than 100 endpoints each.

If we can easily reach more than 100 of instances, then we should try to open an merge request to the OpenTelemetry Helm repository asking the support to podMonitor when not using daemonset.

amit-mazor:
@avivgold098 the reason for servicemonitor is the official otel chart that doesn't allow you to use podmonitor with deployment mode [https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/podmonitor.yaml#L1]

avivgold098 (approved):

nicolastakashi (approved):
lgtm

PR #141, opened 2023-01-12
Add the k8s_events logs receiver as an event exporter

amit-mazor:
Add the `k8s_events logs receiver` as an event exporter. 
In the future we'll add it as a preset to the official otel-chart, but until its a preset, adding it as another regular receiver.
** required permissions are already included here because of the clusterMetrics preset

Changed `defaultSubsystemName` to "metrics"


otel-infrastructure-collector/values.yaml
```diff
@@ -6,7 +6,7 @@ global:
   logs:
     endpoint: "<logs-endpoint-here>"
   defaultApplicationName: "default"
-  defaultSubsystemName: "nodes"
```
avivgold098, line 9: What is the motivation behind?

avivgold098 (commented):
@amit-mazor  can it be as a feature flag? or support it in the upstream?

otel-infrastructure-collector/values.yaml
amit-mazor: the logs of the events-exporter dont include any of the subsystem attributes, so its taking the default subsystem, which is 'nodes' , and its weird to have logs of events under nodes. 
actually metrics is not good since its events logs and not metrics, 
but maybe 'events' or something

nicolastakashi: kube-events maybe.

amit-mazor:
Closed because of the https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver which is best to use instead of `k8s_events` receiver

PR #143, opened 2023-01-19
Add externallabels support to the prometheus template

amit-mazor:
In order to use externalLabels in remoteWrite, we need to be able to specify the external labels , which was not supoprted until now

### DIFF
--- a/metrics/prometheus-agent/override.yaml
+++ b/metrics/prometheus-agent/override.yaml
@@ -1,6 +1,7 @@
 prometheus:
   prometheusSpec:
-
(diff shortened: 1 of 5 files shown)


metrics/prometheus-agent/override.yaml
```diff
@@ -1,6 +1,7 @@
 prometheus:
   prometheusSpec:
-
+    externalLabels:
+      prometheus: coralogix
```
amit-mazor, line 4: its an example

nicolastakashi (approved):

PR #146, opened 2023-01-23
[FEAT] upgrading fluent-bit

nicolastakashi:
I'm upgrading fluent-bit to the latest stable version.

Signed-off-by: Nicolas Takashi <nicolas.takashi@coralogix.com>


amit-mazor:
I think we should also upgrade the fluentbit version in the chart : 
https://github.com/coralogix/telemetry-shippers/blob/master/logs/fluent-bit/k8s-helm/http/values.yaml#L6
[we can do it separately if you prefer, but if we upgrade, lets use it]

amit-mazor (approved):

PR #149, opened 2023-02-01
[FEAT] upgrading fluent-bit helm chart

nicolastakashi:
I'm upgrading the fluent-bit version to 2.0.8 since we have new features like the new internal metrics plugin as described [here](https://docs.fluentbit.io/manual/pipeline/inputs/fluentbit-metrics) and also because this version has available features which can improve the performance on high volume environments such as the threaded mechanism allows input plugins to run in a separate thread which helps to desaturate the main pipeline and move certain loads to a different CPU core.

Because we want to enable the storage metrics by default, we also need to upgrade to the latest helm chart version which allows the configuration of `additionalEndpoints` on the ServiceMonitor and also the VPA manifest of non-enabled features.

Signed-off-by: Nicolas Takashi <nicolas.takashi@coralogix.com>

### DIFF
--- a/logs/fluent-bit/k8s-helm/coralogix/values.yaml
+++ b/logs/fluent-bit/k8s-helm/coralogix/values.yaml
@@ -3,10 +3,13 @@ fluent-bit:
 
   image:
     repository: coralogixrepo/coralogix-fluent-bit-multiarch
-    tag: v0.0.2
+    tag: v0.0.3
 
   serviceMonitor:
     enabled: true
+    additionalEndpoints:
+      - port: metrics
+        path: /metrics
 
   resources:
     limits:
@@ -23,6 +26,12 @@ fluent-bit:
   envFrom:
     - secretRef:
         name: coralogix-keys
+
+  extraPorts:
+    - port: 2021
+      containerPort: 2021
+      protocol: TCP
+      name: metrics
 
   #prometheusRule:
   #  enabled: true
@@ -114,6 +123,7 @@ fluent-bit:
           DB /var/log/fluentbit-tail.db
 
       @INCLUDE input-systemd.conf
+      @INCLUDE input-fluentbit-metrics.conf
 
     filters: |-
       [FILTER]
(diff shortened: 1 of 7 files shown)


doryer:
@nicolastakashi Can you provide if you have notion document which describes the improvements on the version we are upgrading to?

doryer (approved):

logs/fluent-bit/k8s-helm/coralogix/values.yaml
```diff
@@ -3,10 +3,13 @@ fluent-bit:
...
   image:
     repository: coralogixrepo/coralogix-fluent-bit-multiarch
-    tag: v0.0.2
+    tag: v0.0.3
 
   serviceMonitor:
     enabled: true
+    additionalEndpoints:
+      - port: metrics
+        path: /metrics
```
avivgold098, line 12: * Does The upgrade impact the existing `/api/v1/metrics/prometheus`? 
* `/api/v1/metrics/prometheus` will remain the default?

logs/fluent-bit/k8s-helm/coralogix/values.yaml
```diff
@@ -24,6 +27,12 @@ fluent-bit:
     - secretRef:
         name: coralogix-keys
 
+  extraPorts:
+    - port: 2021
+      containerPort: 2021
+      protocol: TCP
+      name: metrics
```
avivgold098, line 34: storage-metrics?

avivgold098 (changes requested):
@nicolastakashi  a few missing from my side - 

*  We need to well explain in the pull request description what is the motivation and mention few highlight features we are going to use - 
 - new storage metrics - will allow better debugging of buffers
* Update the `CHANGELOG.md` 
* Maybe this should not patched version,  this major upgrade
* We need to update in the `Chart.yaml` the - `version` & `appVersion` 


Looking good generally, but important to stick our open source standard

logs/fluent-bit/k8s-helm/coralogix/values.yaml
nicolastakashi: Hi, @avivgold098 yes that path remains the same and there's no impact on the existing metrics, this is something I also discussed with the Fluent-bit upstream charts.

By default the storage metrics are disabled so the default metrics remains the `/api/v2/metrics/prometheus` exporting basic metrics

logs/fluent-bit/k8s-helm/coralogix/values.yaml
nicolastakashi: Agreed!
Can you take a look?

nicolastakashi: Changed!

nicolastakashi:
> @nicolastakashi a few missing from my side -
> 
> * We need to well explain in the pull request description what is the motivation and mention few highlight features we are going to use -
> 
> * new storage metrics - will allow better debugging of buffers
> 
> * Update the `CHANGELOG.md`
> * Maybe this should not patched version,  this major upgrade
> * We need to update in the `Chart.yaml` the - `version` & `appVersion`
> 
> Looking good generally, but important to stick our open source standard

I added new updated to the change log and also to the PR description, can you take a look again?

avivgold098 (approved):
Looking good :)

PR #156, opened 2023-02-13
[UPDATE] otel helmchart version

nicolastakashi:
I'm upgrading OTEL helm chart version to be able to use the latest OTEL version and leverage k8sobjectreceiver

### DIFF
--- a/otel-agent/CHANGELOG.md
+++ b/otel-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## OpenTelemtry-Agent
 
+### v0.0.20 / 2023-02-13
+
+* [UPRADE] Upgrading chart version from 0.47.0 to 0.48.1
+
 ### v0.0.19 / 2023-01-26
 
 * [CHORE] Update OpenTelemetry Collector to v0.70.0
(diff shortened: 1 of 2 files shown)


amit-mazor:
The `k8sobjectsreceiver receiver` is supported since version v0.63.0 and we are currently in version 0.70.0, is there a difference between 0.70.0 and 0.71.0 [in this receiver] ? 

Anyways Im in favor of upgrading, we just need to make sure there won't be any `breaking changes` because I see in the release notes under breaking changes both `spanmetricsprocessor` and `exporter/coralogix` [we use both]

amit-mazor (approved):

nicolastakashi:
> The `k8sobjectsreceiver receiver` is supported since version v0.63.0 and we are currently in version 0.70.0, is there a difference between 0.70.0 and 0.71.0 [in this receiver] ?
> 
> Anyways Im in favor of upgrading, we just need to make sure there won't be any `breaking changes` because I see in the release notes under breaking changes both `spanmetricsprocessor` and `exporter/coralogix` [we use both]

Since we're already using version 0.70, there're no breaking change on this upgrade.

PR #169, opened 2023-02-27
Update README.md

juan-coralogix:
Please take a look if the explanation is better.  Any suggestion is more than welcome.


amit-mazor:
@juan-coralogix Im not really in favor of adding this operator installation guide here because its not part of prometheus agent itself and its too much information that is not related to the agent itself. 

We already have [a prometheus operator installation page](https://github.com/coralogix/telemetry-shippers/tree/master/metrics/prometheus/operator) which installs the operator with Helm. 

If we want to suggest another option for installing the operator , with bundle as you suggested, we can mention it under the prometheus operator readme, not here.

Anyways - I think its great to add these examples of podmonitor and how to use it, but lets add it under the prometheus operator page :)

PR #170, opened 2023-02-27
Update fluentd-image.yml

MichaelBriggs-Coralogix:
Add workflow_dispatch option
Remove "latest" tag

### DIFF
--- a/.github/workflows/fluentd-image.yml
+++ b/.github/workflows/fluentd-image.yml
@@ -6,6 +6,7 @@ on:
     branches: master
     paths:
     - 'logs/fluentd/image/**'
+  workflow_dispatch:
 
 env:
   DOCKERHUB_REGISTRY: coralogixrepo/coralogix-fluentd-multiarch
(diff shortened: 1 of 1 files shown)


.github/workflows/fluentd-image.yml
```diff
@@ -40,6 +41,4 @@ jobs:
           context: ./logs/fluentd/image
           platforms: linux/amd64,linux/arm64
           push: true
-          tags: |
-            ${{ env.DOCKERHUB_REGISTRY }}:${{ steps.get-fluentd-version.outputs.version }}
-            ${{ env.DOCKERHUB_REGISTRY }}:latest
```
daidokoro, line 45: Hey @MichaelBriggs-Coralogix ,

Do we want to remove the override for the latest tag? It what allow the user to get the most recent version when they pull image:latest.

MichaelBriggs-Coralogix: Per my understanding, we're wanting to remove all "latest" tags from our Docker images (at least for our integrations) to ensure we explicitly set them in our charts. This is to ensure stability/supportability.

daidokoro: Got it.

daidokoro (approved):
LGTM

PR #172, opened 2023-03-03
Fix for OTEL Agent k8s Helm Chart docs

itsmegb:
The port number used in the example code for the OTEL_EXPORTER_OTLP_ENDPOINT was incorrect and would cause an instrumented pod to be unable to send traces to OTEL.

I have updated the port number in the documentation and also made some minor clean-up changes to the markdown (not yaml !) in the readme.


matej-g (approved):
Thank you @itsmegb! Looks good to me.

PR #173, opened 2023-03-06
[FEATURE] changing prometheus agent to use default configs

nicolastakashi:
I'm changing the Prometheus agent to use default values in terms of shards and max samples sent and also adding a self-monitor feature as well as a snapshot on WAL to improve restarts.

### DIFF
--- a/metrics/prometheus-agent/values.yaml
+++ b/metrics/prometheus-agent/values.yaml
@@ -6,6 +6,10 @@ global:
 
 prometheus:
   prometheusSpec:
+    routePrefix: /
+    enableFeatures:
+      - memory-snapshot-on-shutdown
+      - new-service-discovery-manager
     image: 
       repository: quay.io/prometheus/prometheus
       tag: v2.42.0
@@ -21,7 +25,7 @@ prometheus:
       name: prometheus-agent-coralogix
       queueConfig:
         capacity: 2500
-        maxSamplesPerSend: 1000
+        maxSamplesPerSend: 500
         maxShards: 200
       remoteTimeout: 120s
       url: https://prometheus-gateway.eu2.coralogix.com/prometheus/api/v1/write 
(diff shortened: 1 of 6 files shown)


metrics/prometheus-agent/values.yaml
```diff
@@ -62,3 +66,27 @@ prometheus:
...
+
+    scheme: ""
+
+    tlsConfig: {}
+
+    bearerTokenFile:
+
+    metricRelabelings: []
+    
+    relabelings: []
```
amit-mazor, line 91: since these whole configs are related to servicemonitor , i would remove the extra lines between so it wont be confusing and be clear its all servicemonitor configs

metrics/prometheus-agent/values.yaml
```diff
@@ -21,7 +25,7 @@ prometheus:
       name: prometheus-agent-coralogix
       queueConfig:
         capacity: 2500
-        maxSamplesPerSend: 1000
+        maxSamplesPerSend: 500
```
amit-mazor, line 28: if we lower it, how will it affect the remote write ? 
will it require less resources, will it be slower ?

nicolastakashi: It will depend on the amount of data shipped, since this is a repository we're offering to our customers, we should not define how many samples they are going to send or how many shards they need to have, I'm changing by default values we have on Prometheus.

At Coralogix we're going to figure out what's the proper value for our use case.

metrics/prometheus-agent/values.yaml
nicolastakashi: Done :D

metrics/prometheus-agent/values.yaml
amit-mazor: i understand, agree

amit-mazor:
LGTM

amit-mazor (approved):

PR #174, opened 2023-03-07
[CHORE] adding prom agent svc

nicolastakashi:
I'm adding the missing Prometheus Agent Service


metrics/prometheus-agent/templates/prometheus.yaml
```diff
@@ -8,7 +8,6 @@ metadata:
     app.kubernetes.io/name: {{ template "prometheus-agent.name" . }}
     app.kubernetes.io/component: prometheus
     release: {{ $.Release.Name | quote }}
-    self-monitor: {{ .Values.prometheus.serviceMonitor.selfMonitor | quote}}
```
amit-mazor, line 11: why do we remove it. ?

nicolastakashi: Because I added by mistake on the previous PR

amit-mazor (approved):

PR #177, opened 2023-03-13
Update README.md

juan-coralogix:
Update Main Readme warning for 1.25+ K8s podSecurity

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,14 @@ Please see [#Getting Started](README.md#getting-started) for more information ab
 This repository contains directories for each integration type, logs and metrics, and open-telemetry that can send all.  
 Inside each integration type there are multiple ways to install our integrations, using helm, installing kubernetes manifests directly, using a docker image or installing a service.
 
+## Kubernetes 1.25+
+podsecuritypolicy has been deprecated for this version up.
+You can disable this by adding the following to your override file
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,6 +12,14 @@ Please see [#Getting Started](README.md#getting-started) for more information ab
 This repository contains directories for each integration type, logs and metrics, and open-telemetry that can send all.  
 Inside each integration type there are multiple ways to install our integrations, using helm, installing kubernetes manifests directly, using a docker image or installing a service.
 
+## Kubernetes 1.25+
+podsecuritypolicy has been deprecated for this version up.
+You can disable this by adding the following to your override file
+```bash
+podSecurityPolicy:
+  create: false
```
amit-mazor, line 20: enabled: false [not create]

amit-mazor:
I would move this to Fluentd chart README , and not add it to the main readme , because its not clear what its related to, and in fluentbit for example its already disabled by default so not relevant there. 

(soon I will disable it also for fluentd so we will remove this note, unless its relevant in other places that are not fluent)

amit-mazor:
I pushed this change , and added a note in the readme , so we can close this PR 
https://github.com/coralogix/telemetry-shippers/pull/178/files

PR #181, opened 2023-03-20
[FEATURE] collecting kubernetes events

nicolastakashi:
I'm adding a new logs pipeline to collect Kubernetes events and export than to Coralogix.

### DIFF
--- a/otel-infrastructure-collector/k8s-helm/values.yaml
+++ b/otel-infrastructure-collector/k8s-helm/values.yaml
@@ -13,6 +13,11 @@ opentelemetry-collector:
   fullnameOverride: otel-infrastructure-collector
   clusterRole:
     name: "otel-infrastructure-collector"
+    create: true
+    rules: 
+    - apiGroups: ["", "events.k8s.io"]
+      resources: ["events"]
+      verbs: ["watch", "list"]
     clusterRoleBinding:
       name: "otel-infrastructure-collector"
   replicaCount: 1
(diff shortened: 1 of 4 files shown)


otel-infrastructure-collector/k8s-helm/values.yaml
```diff
@@ -13,6 +13,11 @@ opentelemetry-collector:
   fullnameOverride: otel-infrastructure-collector
   clusterRole:
     name: "otel-infrastructure-collector"
+    create: true
+    rules: 
+    - apiGroups: ["", "events.k8s.io"]
+      resources: ["events"]
+      verbs: ["watch", "list"]
```
amit-mazor, line 20: because we enable the `clusterMetrics` preset , it creates a cluster role with watch and list on events resource, for "" apigroup, isnt that enough ? see: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/clusterrole.yaml#L23

nicolastakashi: Users can disable clusterMetrics as we're doing internally :)

amit-mazor: but if they dont ,  they have 2 clusterroles ?

amit-mazor: right , and i see the condition is with 'or' , so it wont be duplicated , great!

amit-mazor (approved):

amit-mazor:
LGTM

PR #184, opened 2023-03-21
Update default otel configuration

daidokoro:
Adds resource detection for Docker attributes to Otel Agent default config for metrics/traces

### DIFF
--- a/otel-agent/ecs-ec2/config.yaml
+++ b/otel-agent/ecs-ec2/config.yaml
@@ -32,8 +32,8 @@ processors:
   batch:
   batch/metrics:
     timeout: 60s
-  resourcedetection/ecs:
-    detectors: [env, ecs]
+  resourcedetection:
+    detectors: [env, ec2, ecs, docker, system]
     timeout: 2s
     override: false
 exporters:
@@ -62,7 +62,7 @@ service:
         - logging
       processors:
         - batch
-        - resourcedetection/ecs
+        - resourcedetection
       receivers:
         - otlp
     metrics:
@@ -71,7 +71,7 @@ service:
         - logging
       processors:
         - batch
-        - resourcedetection/ecs
+        - resourcedetection
       receivers:
         - hostmetrics
         - prometheus


royfur (approved):

otel-agent/ecs-ec2/config.yaml
```diff
@@ -32,8 +32,8 @@ processors:
   batch:
   batch/metrics:
     timeout: 60s
-  resourcedetection/ecs:
-    detectors: [env, ecs]
+  resourcedetection:
+    detectors: [env, ec2, ecs, docker, system]
```
matej-g, line 36: Do we want to use both `docker` and `system`? As far as I understood the [docs](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/README.md#system-metadata), both give you the same info (host name and OS), but you should use Docker if running collector as Docker container.

matej-g (commented):
Looks good 👍, small question about using both `system` and `docker`

otel-agent/ecs-ec2/config.yaml
daidokoro: Not sure why but in testing, when I removed system, the **host.id** would be missing from the resource attributes. I can check the further to be sure.

matej-g: I guess then we could use `system` only? Not sure how important / usable the `host.id` information is when running in ECS. Docker does not provide this one I think.

daidokoro: I believe host.id and host.name, etc are required for APM. Also note removing `docker` will also make the the `host.id` and other attributes disappear.

The otel docs on this seem misleading.

Attributes **without** `system`:
Screenshot 2023-03-22 at 11 21 34

Attributes **with** `system`
Screenshot 2023-03-22 at 11 43 42

matej-g: Thanks for the detailed explanation! Let's keep it then.

PR #190, opened 2023-03-31
[FIX] Prometheus Agent logs configuration

nicolastakashi:
Since we're creating our own Prometheus CRD, because Prometheus Operator still doesn't have Prometheus Agent Available, we need to change the logs config from the CRD Properties to the container flags

### DIFF
--- a/metrics/prometheus-agent/templates/prometheus.yaml
+++ b/metrics/prometheus-agent/templates/prometheus.yaml
@@ -29,6 +29,8 @@ spec:
     - --web.console.libraries=/etc/prometheus/console_libraries
     - --web.route-prefix=/
     - --web.config.file=/etc/prometheus/web_config/web-config.yaml
+    - --log.level={{ .Values.prometheus.prometheusSpec.logLevel }}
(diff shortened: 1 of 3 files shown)


metrics/prometheus-agent/templates/prometheus.yaml
```diff
@@ -29,6 +29,8 @@ spec:
     - --web.console.libraries=/etc/prometheus/console_libraries
     - --web.route-prefix=/
     - --web.config.file=/etc/prometheus/web_config/web-config.yaml
+    - --log.level={{ .Values.prometheus.prometheusSpec.logLevel }}
+    - --log.format={{ .Values.prometheus.prometheusSpec.logFormat }}
```
oded-dd, line 33: Should default to json (I would even set it up as json hardcoded)

nicolastakashi: This is json by default and also warn by default, but since it's a public chart I prefer to let the option to change available.

oded-dd (approved):

PR #199, opened 2023-05-02
[otel-agent] allow users to configure coralogix domain

povilasv:
This change shouldn't break users since if old values are configured they take priority over new config.

But in case the endpoint values are empty (new default) we use the value from `domain` field


otel-agent/k8s-helm/Chart.yaml
```diff
@@ -8,7 +8,7 @@ keywords:
   - Coralogix
 dependencies:
   - name: opentelemetry-collector
-    version: "0.52.0"
+    version: "0.55.0"
```
matej-g, line 11: It looks like the Domain enhancement was released only in `0.76.3` collector (https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.76.3).

Should we first bump the version in the Helm chart, since it's `0.76.1` right now?

nicolastakashi:
Should we do the same for otel-infrastructure?

otel-agent/k8s-helm/Chart.yaml
matej-g: @povilasv has directed me to this comment https://github.com/open-telemetry/opentelemetry-helm-charts/pull/772#issuecomment-1531564324 explaining the versioning situation, looks like `0.76.1` is the correct version to use. Resolving this thread 👍.

matej-g (approved):
Looks good ✅ . We might need to merge latest `master` again, looks like with the docs check now there are still couple of dead links causing it to fail.

nicolastakashi (approved):

ArthurSens:
We also need to update the docker-compose example :)

[Brian hit](https://coralogix-dev.slack.com/archives/C02SLSEP2CS/p1683125428387939) this problem today 😅

PR #209, opened 2023-05-04
Add resourcedetection to logging.yaml

MichaelBriggs-Coralogix:


oded-dd (commented):
I think it makes more sense to add it as presets, as it has network traffic costs. @povilasv what do you think?

nicolastakashi:
> @oded-dd this is just example configuration and since usually in all our integration we enrich with similar data we wanted to have it as well in this example at the end customer can decide if to remove it or not

@royfur you're adding to the `otel-agent` this means that everybody is using this chart is going to get this config out-of-the-box, Oded's suggestion is to have this trough a preset which is disabled by default and could be enabled through a flag.

What I also agree btw.

MichaelBriggs-Coralogix:
@oded-dd and @nicolastakashi 
Unless I'm mistaken, the otel-agent is not consuming the example logging.yaml file, but is instead consuming config.yaml. Additionally, the resourcedetection processor that I proposed gets added to logging.yaml is already present and enabled in config.yaml by default.

royfur:
> > @oded-dd this is just example configuration and since usually in all our integration we enrich with similar data we wanted to have it as well in this example at the end customer can decide if to remove it or not
> 
> @royfur you're adding to the `otel-agent` this means that everybody is using this chart is going to get this config out-of-the-box, Oded's suggestion is to have this trough a preset which is disabled by default and could be enabled through a flag.
> 
> What I also agree btw.

this is not adding to otel-agent default config, it is to the ecs-ec2 deployment and it is always relevant there

royfur (approved):

PR #211, opened 2023-05-12
[otel-agent] add openshift support

povilasv:


otel-agent/k8s-helm/values.yaml
```diff
@@ -2,6 +2,7 @@ global:
   domain: ""
   defaultApplicationName: "default"
   defaultSubsystemName: "nodes"
+  fullnameOverride: otel-coralogix
```
matej-g, line 5: Do we need both `fullnameOverride`?

otel-agent/k8s-helm/templates/scc.yaml
```diff
@@ -0,0 +1,49 @@
+{{/*
+Default values for SecurityContextConstraints
+*/}}
+
+{{- define "opentelemetry-coralogix.defaultSecurityContextConstraints" -}}
+priority: 10
+allowHostNetwork: true
+allowHostPorts: true
+allowHostPID: true
+allowHostDirVolumePlugin: true
```
matej-g, line 10: Do we know if all of these need to be allowed? Or could we at least put a comment why is each needed? AFAIK we need host network and host dir, but since I'm unsure what each one does, I cannot tell if we really need any other besides the two.

nicolastakashi:
LGTM 😄 
I guess we should just add an entry on the change log.

WDYT?

matej-g (commented):
Looking good! 🙌 Few small ones.

matej-g (approved):

PR #226, opened 2023-05-22
fix default values for prometheus.monitoring.coreos.com/v1

adam-oligo:
# Description

Providing null values is not allowed for this object thus I provided default values as per the [official prometheus community repo](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml)

error received when trying to apply:
```
Prometheus.monitoring.coreos.com "prometheus-agent" is invalid: [spec.disableCompaction: Invalid value: "null": spec.disableCompaction in body must be of type boolean: "null", spec.retention: Invalid value: "null": spec.retention in body must be of type string: "null", spec.alerting: Invalid value: "null": spec.alerting in body must be of type object: "null", spec.ruleNamespaceSelector: Invalid value: "null": spec.ruleNamespaceSelector in body must be of type object: "null", spec.walCompression: Invalid value: "null": spec.walCompression in body must be of type boolean: "null", spec.ruleSelector: Invalid value: "null": spec.ruleSelector in body must be of type object: "null"]
```

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/metrics/prometheus-agent/templates/prometheus.yaml
+++ b/metrics/prometheus-agent/templates/prometheus.yaml
@@ -14,7 +14,7 @@ spec:
   ruleSelector: {}
   walCompression: true
   disableCompaction: false
-  retention: 10d
+  retention: 2h
   containers:
   - args:
     - --config.file=/etc/prometheus/config_out/prometheus.env.yaml
--- a/metrics/prometheus-agent/templates/prometheus.yaml
+++ b/metrics/prometheus-agent/templates/prometheus.yaml
@@ -1,7 +1,7 @@
 apiVersion: monitoring.coreos.com/v1
 kind: Prometheus
 metadata:
-  name: {{ template "prometheus-agent.name" . }} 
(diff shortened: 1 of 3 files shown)


metrics/prometheus-agent/templates/prometheus.yaml
```diff
@@ -1,20 +1,20 @@
...
-  ruleSelector: null
-  walCompression: null
-  disableCompaction: null
-  retention: null
+  alerting: {}
+  ruleNamespaceSelector: {}
+  ruleSelector: {}
+  walCompression: true
+  disableCompaction: false
+  retention: 10d
```
oded-dd, line 17: Why 10days? This is agent. It should not keep data at all (only the first 2 hours).

adam-oligo: no probs it was the default values, I will change it

adam-oligo:
> LGTM. Thank you :)

How long do you think this will be merged? It's block our sync

PR #228, opened 2023-05-22
fix issue with prometheus agent and operator

adam-oligo:
# Description

Since upgrading to kube-prometheus-stack 45.30.0 there is an issue with the chart which causes empty labels to always be printed in prometheus.rules, for example [here](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus/rules-1.14/node-exporter.yaml#L47)

this will be rendered as follows:
image

adding new values fixes the issue and renders the rules correcly.
I couldn't find a valid reason for this to happen so had to use this workaround.

In addition removed non required values from prometheus-agent which caused it crash
 Fixes #

# How Has This Been Tested?

deployed locally

# Checklist:

### DIFF
--- a/metrics/prometheus-agent/CHANGELOG.md
+++ b/metrics/prometheus-agent/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Prometheus-Agent
 
+### v0.0.14 / 2023-05-22
+* [FIX] Remove non required objects from Prometheus object
+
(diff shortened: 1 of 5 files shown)


oded-dd (approved):

adam-oligo:
@oded-dd merge please :)

PR #231, opened 2023-05-23
[k8s-helm-windows] fix bugs and add Service

povilasv:
# Description

[SC-18267]

* Add Service for DaemonSet, since windows does not support hostnetworking
* Fix includeCollectorLogs=true breaks configuration.
* Add support for Deployment and StatefulSet options.

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/otel-agent/k8s-helm-windows/Chart.yaml
+++ b/otel-agent/k8s-helm-windows/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-coralogix-windows
-version: 0.0.1
+version: 0.0.2
 description: Windows OpenTelemetry agent to which instrumentation libraries export their telemetry data
 type: application
 appVersion: 0.77.0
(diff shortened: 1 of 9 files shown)


otel-agent/k8s-helm-windows/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-coralogix-windows
-version: 0.0.1
+version: 0.0.2
```
matej-g, line 3: I guess we'll need to add another CI check for the new chart, but I'm happy to do it as a follow up once we also have images published.

otel-agent/k8s-helm-windows/templates/service.yaml
```diff
@@ -27,7 +26,7 @@ spec:
     {{- end }}
   {{- end }}
   ports: {{ include "opentelemetry-collector.deploymentPortsConfig" . | nindent 4 }}
+  internalTrafficPolicy: {{ include "opentelemetry-collector.serviceInternalTrafficPolicy" . }}
```
matej-g, line 29: Is this change Windows related / required for Windows?

matej-g (approved):
Looks good to me 👍, just one clarification.

oded-dd (approved):

PR #232, opened 2023-05-23
Add OpenTelemetry Collector Contrib images for Windows

matej-g:
# Description
[SC-18267]

This PR adds set up and workflow to build and publish collector (contrib distribution) images for Windows to our repository. This is an intermittent solution before there are official Docker images in the upstream.

We build two sets of images, both for particular Windows server versions (2019 and 2022).

# How Has This Been Tested?
Images have been tested on an EKS cluster.

# Checklist:


matej-g:
FYI that failing link is legit, it's because we need to fill it after we publish the images (it's just a todo). I'll remove it for now.

oded-dd (approved):
LGTM

PR #242, opened 2023-06-21
Added example on CRI parsing longlogs

juan-coralogix:
# Description

Just adds an example to Readme.
On how to fix LongLogs in CRI / containerd with concat.

# How Has This Been Tested?

Its a readme, no testing done. But the configuration added to the readme has been tested in a UseCase scenario.
# Checklist:
- [X] This change does not affect any particular component (e.g. it's CI or docs change)


nicolastakashi:
@juan-coralogix can you run `make fmt-docs` and push?

logs/fluentd/k8s-helm/http/README.md
```diff
@@ -88,3 +88,32 @@ By default this chart installs additional dependent chart:
 ## Coralogix Fluentd Buffer Alert
 
 In order to create an alert on Fluentd buffer in Coralogix, please see [coralogix-alert doc](https://github.com/coralogix/telemetry-shippers/blob/master/logs/fluentd/docs/coralogix-alerts.md)
+
+
+## Log Logs: containerd / CRI partial logs
+
+If your application is generating logs longer than 16k you should notice that docker dirver is splitting the log in multiple messages.
```
matej-g, line 95:
```suggestion
If your application is generating logs longer than 16k, you should note that the Docker driver will split the logs into multiple messages.
```

logs/fluentd/k8s-helm/http/README.md
```diff
@@ -88,3 +88,32 @@ By default this chart installs additional dependent chart:
...
+To fix this we can use concat to fix this.
```
matej-g, line 96:
```suggestion
To resolve this, you can use the `concat` filter.
```

logs/fluentd/k8s-helm/http/README.md
```diff
@@ -88,3 +88,32 @@ By default this chart installs additional dependent chart:
...
+
+First lets make sure that in the override file, that you use to deploy the helm, has logtag as one of the regex group keys, just like this.
```
matej-g, line 98:
```suggestion
First ensure that in the override file used to deploy the Helm chart has `logtag` as one of the regex group keys, like the following:
```

logs/fluentd/k8s-helm/http/README.md
```diff
@@ -88,3 +88,32 @@ By default this chart installs additional dependent chart:
...
+```yaml
+<pattern>
+  format /^(?<time>[^ ]+) (?<stream>stdout|stderr) (?<logtag>[^ ]*) (?<message>.*)$/
+  time_format %Y-%m-%dT%H:%M:%S.%L%z
+  keep_time_key true
+</pattern>
+```
+If that is not the case please replace the existing one with this one.
+
+Then next to the source we will add the following filter that will concat the logs:
```
matej-g, line 109:
```suggestion
Then, next to the source, you'll need to add the following filter that will concat the logs:
```

matej-g (commented):
Looks good, some stylistic suggestions + @nicolastakashi already mentioned we should fix the failing CI

nicolastakashi (approved):

PR #249, opened 2023-07-11
[otel-agent] add helm install test

povilasv:
# Description

Adds smoke test -  helm install test , which tries to install otel-agent

 Fixes #ES-45

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/.github/workflows/helm-test.yml
+++ b/.github/workflows/helm-test.yml
@@ -1,0 +1,20 @@
+name: Otel Agent Helm Install Test
+
+on:
+  pull_request:
+
+jobs:
+  collector-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/helm-test.yml
```diff
@@ -0,0 +1,20 @@
...
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Setup
+        uses: ./.github/actions/setup
+        with:
+          create-kind-cluster: "true"
+      - name: Run chart-testing (install)
+        run: ct install --charts otel-agent/k8s-helm
```
matej-g, line 19: I assume this strictly only checks if the Helm installation succeeds?

matej-g (approved):
This looks like a good basis to me, but it would be nice if we could add (even in later PRs) some extra steps to validate that the agent is actually running and healthy.

PR #258, opened 2023-07-24
Introduce OpenTelemetry Integration chart

matej-g:
# Description

Resolves [ES-55](https://coralogix.atlassian.net/browse/ES-55).

This PR introduces a new chart called OpenTelemetry Integration part, that combines functionalities of both agent and the infrastructure collector (+ collects metrics for K8s dashboard out of the box). Includes as well CI changes and detailed documentation. For more details see the ticket description.

# How Has This Been Tested?

# Checklist:

[ES-55]: https://coralogix.atlassian.net/browse/ES-55?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


otel-integration/k8s-helm/values.yaml
```diff
@@ -0,0 +1,470 @@
...
+      - health_check
+      - memory_ballast
+      pipelines:
+        metrics:
+          exporters:
+            - coralogix
+          processors:
+            - k8sattributes
+            - resourcedetection/env
+            # - resourcedetection/region
```
nicolastakashi, line 200: Should we remove the comment?

otel-integration/k8s-helm/values.yaml
```diff
@@ -0,0 +1,470 @@
...
+      memory: 2G
+
+  ports:
+    jaeger-binary:
+      enabled: true
+      containerPort: 6832
+      servicePort: 6832
+      hostPort: 6832
+      protocol: TCP
+    # In order to enable podMonitor, following part must be enabled in order to expose the required port:
```
nicolastakashi, line 245: Sorry for the dummy question, but why this is needed?

otel-integration/k8s-helm/values.yaml
```diff
@@ -0,0 +1,470 @@
...
+
+  # podMonitor:
+  #   enabled: true
+
+  # prometheusRule:
+  #   enabled: true
+  #   defaultRules:
+  #     enabled: true
+
+opentelemetry-collector-infrastucture:
```
nicolastakashi, line 257: Just sharing a feeling!
I know naming is hard but I don't like the infrastructure suffix, tbh I don't have anything better to suggest. 😢

otel-integration/k8s-helm/values.yaml
```diff
@@ -0,0 +1,470 @@
...
+      - memory_ballast
+      pipelines:
+        metrics:
+          exporters:
+            - coralogix
+          processors:
+            - k8sattributes
+            - resourcedetection/env
+            # - resourcedetection/region
+            - metricstransform
```
nicolastakashi, line 201: On the monitoring shippers we are ensuring that we will not collect more than is needed for the K8S Dashboard work.

https://github.com/coralogix/monitoring-shippers/blob/main/charts/values.yaml#L146

Should we keep this behavior here?

.github/workflows/otel-integration.yml
```diff
@@ -0,0 +1,41 @@
...
+          helm package .
+      -
+        name: Setup JFrog CLI
+        uses: jfrog/setup-jfrog-cli@v2.1.0
+        with:
+          version: 2.12.1
+      -
+        name: use-jfrog-cli
+        run: |
+          cd otel-agent/k8s-helm
```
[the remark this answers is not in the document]
matej-g: You're right 🤦 , fixing it.

otel-integration/k8s-helm/values.yaml
matej-g: Good eyes. Looks like the comment got lost, I added a TODO here - we first need to upgrade the collector, which is blocked by https://coralogix.atlassian.net/browse/ES-51, since we need latest feature for some detectors and configs to work. I'll re-add the comment or ideally try to unblock before this is merge.

otel-integration/k8s-helm/values.yaml
matej-g: Dummy answer: copy-paste 🙈.

Real answer: Are you referring to `ports` section or this pod monitor config?

otel-integration/k8s-helm/values.yaml
matej-g: I feel ya, I'm also not totally sold on the name 😢. Mostly just took it over from existing chart. Would something like `opentelemetry-cluster-collector` make sense, or `opentelemetry-k8s-collector`, since we're collecting metrics and events related to the cluster? Not sure...

otel-integration/k8s-helm/values.yaml
matej-g: So it might be a bit confusing here, but this is now:
- For cluster receiver metrics, we don't filter out, since if `clusterMetrics` preset is enabled, we send metrics from that receiver (based on the preset config)
- For the extra metrics (KSM and API / node metrics), these are filtered in the chart as part of the `kubernetesExtraMetrics` preset (see [here](https://github.com/coralogix/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/_config.tpl#L195) and [here](https://github.com/coralogix/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/_config.tpl#L378))

We perhaps could do some more complex checking (i.e. when K8s dashboard extension is enabled, but cluster metrics disabled, we would only collect necessary metrics for dashboard), but I did not find a good way of doing this 😒. So for now it is kind of assumed that we collect all cluster receiver metrics, if this preset is enabled, regardless of the k8s dashboard parameter. I hope that works.

otel-integration/k8s-helm/values.yaml
nicolastakashi: The pod monitor one hehe 😄

otel-integration/k8s-helm/values.yaml
nicolastakashi: I like `opentelemetry-cluster-collector` because it might run not only on k8s, but also openshift 😄

otel-integration/k8s-helm/values.yaml
nicolastakashi: I guess this is Ok

otel-integration/k8s-helm/values.yaml
matej-g: Let's do it 🎉

otel-integration/k8s-helm/values.yaml
matej-g: I guess this was taken over from the upstream for users who might want to enable it, so they can just uncomment it? Not 100% sure, it was really taken over from existing `otel-agent` / `otel-infrastructure-collector` configs.

nicolastakashi (approved):

PR #264, opened 2023-08-03
[otel-integration] Add allocatable metrics; drop unused version attribute

matej-g:
# Description

Reflects some changes done in `monitoring-shipper`.

- Add some extra metrics regarding allocatable resource on cluster
- Remove unused integration version attribute to lower cardinality

# How Has This Been Tested?

Locally.

# Checklist:

### DIFF
--- a/otel-integration/k8s-helm/values.yaml
+++ b/otel-integration/k8s-helm/values.yaml
@@ -1,6 +1,6 @@
 global:
-  domain: ""
-  clusterName: ""
+  domain: "coralogix.com"
+  clusterName: "test"
   defaultApplicationName: "otel"
   defaultSubsystemName: "integration"
   logLevel: "warn"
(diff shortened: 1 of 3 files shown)


otel-integration/k8s-helm/values.yaml
```diff
@@ -457,7 +455,6 @@ opentelemetry-cluster-collector:
   #     enabled: true
 
 kube-state-metrics:
-  enabled: true
```
nicolastakashi, line 460: Is it supposed?

matej-g: Yes, I added line in the changelog to explain. We enable KSM based on if `extensions.kubernetesDashboard.enabled` is set in the global values (https://github.com/coralogix/telemetry-shippers/blob/master/otel-integration/k8s-helm/Chart.yaml#L25). This parameter was created initially but it's not used for anything actually.

PR #270, opened 2023-08-11
[CHORE] upgrading upstream helm chart

nicolastakashi:
# Description

Upgrading the upstream chart to use the latest coralogix otel helm chart

 

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/otel-agent/k8s-helm/CHANGELOG.md
+++ b/otel-agent/k8s-helm/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## OpenTelemtry-Agent
 
+### v0.0.30 / 2023-08-11
+* [CHORE] Upgrading upstream chart. (v0.69.0)
+
 ### v0.0.29 / 2023-07-26
 
 * [FEATURE] Add system.memory.utilization and system.cpu.utilization to hostmetrics preset. (v0.64.1)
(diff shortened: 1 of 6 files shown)


otel-agent/k8s-helm/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## OpenTelemtry-Agent
 
+### v0.0.30 / 2023-08-11
+* [FEATURE] Upgrading upstream chart. (v0.69.0)
+
```
matej-g, line 7: Very important question: Feature or chore? 😁 

```suggestion
### v0.0.30 / 2023-08-11

* [FEATURE] Upgrading upstream chart. (v0.69.0)

```

otel-infrastructure-collector/k8s-helm/Chart.yaml
```diff
@@ -1,15 +1,15 @@
...
 keywords:
   - OpenTelemetry Collector
   - OpenTelemetry Infrastructure Collector
   - Coralogix
 dependencies:
   - name: opentelemetry-collector
-    version: "0.55.0"
-    repository: https://open-telemetry.github.io/opentelemetry-helm-charts
+    version: "0.69.0"
+    repository: https://cgx.jfrog.io/artifactory/coralogix-charts-virtual
```
matej-g, line 12: I wonder if there's anything changing the upstream chart to look out for, but I'm not sure. Probably if the smoke tests pass we should be safe.

matej-g (commented):
~~Looks good, would you mind also updating the `otel-integration` chart? 🙏 Or prefer to do it separately?~~

Actually scratch that, since we have now `0.70.0` I will need to bump it for another feature, so no need.

PR #280, opened 2023-08-21
[otel-integration] feat support host.id from system resource detector

povilasv:
# Description

For bare-metal use case we don't have `host.id`, which is required for k8s dashboard. The host.id is only coming from gcp / aws detectors. This PR changes to use the `system` detector's host.id detection.

Fixes #ES-70

# How Has This Been Tested?

- installed on kind cluster

# Checklist:


matej-g (approved):
Nicely done 🙇

PR #288, opened 2023-09-13
[otel-agent] allow setting cluster name

povilasv:
# Description

Allow users to easily set clusterName.

Tested with kind cluster:
```
helm install --values new-values.yaml otel-agent .  
```

My example generated:
```
        resource/metadata:
          attributes:
          - action: upsert
            key: k8s.cluster.name
            value: 'kind'
          - action: upsert
            key: cx.otel_integration.name
            value: coralogix-otel-agent-helm
```

Fixes https://coralogix.atlassian.net/browse/ES-21

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/otel-agent/k8s-helm/values.yaml
+++ b/otel-agent/k8s-helm/values.yaml
@@ -1,5 +1,6 @@
 global:
   domain: ""
+  clusterName: ""
   defaultApplicationName: "default"
   defaultSubsystemName: "nodes"
   fullnameOverride: otel-coralogix
@@ -48,6 +49,10 @@ opentelemetry-collector:
       enabled: true
     kubeletMetrics:
       enabled: true
+    metadata:
+      enabled: true
+      clusterName: "{{.Values.global.clusterName}}"
+      integrationName: "coralogix-otel-agent-helm"
 
   extraEnvs:
   - name: CORALOGIX_PRIVATE_KEY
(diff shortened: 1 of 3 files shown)


otel-agent/k8s-helm/values.yaml
```diff
@@ -48,6 +49,10 @@ opentelemetry-collector:
       enabled: true
     kubeletMetrics:
       enabled: true
+    metadata:
+      enabled: true
+      clusterName: "{{.Values.global.clusterName}}"
+      integrationName: "coralogix-otel-agent-helm"
```
matej-g, line 55: We use this parameter / attribute in `otel-integration` to identify metrics for the K8s dashboard. Does the same apply for this chart? Or is there other reason to introduce it?

matej-g: Not strictly against, I'm just thinking if we have a good enough reason to have this at the moment, as for `otel-integration` it is required for proper functioning of dashboards. Just trying to decrease the number of attributes, to keep cardinality / costs low for users, given this will be attached to every metric.

matej-g: Maybe not directly but I still think it's a good rule to try keep number of labels to minimum, unless we do believe these are useful. Let's keep it then if we see value in having them.

matej-g (approved):

PR #294, opened 2023-09-29
[FIX] Remove redundant `-agent` from `fullnameOverride` field of `values.yaml`

Vizzyy:
# Description

The `fullnameOverride` in the Values file has a redundant "-agent" at the end. When following the default installation pattern this yields pods that have names like `coralogix-opentelemetry-agent-agent-xxxxx`. 

# How Has This Been Tested?

Yes, I deployed the default configuration, as-is, and you will see it as the attache screenshot. After deleting everything and trying it again without the redundant "-agent" everything is name correctly. No functional difference between either scenario. 

# Checklist:
Screenshot 2023-09-29 at 1 55 30 PM

### DIFF
--- a/otel-integration/CHANGELOG.md
+++ b/otel-integration/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## OpenTelemtry-Integration
 
+### v0.0.17 / 2023-09-29
+
+* [FIX] Remove redundant `-agent` from `fullnameOverride` field of `values.yaml`
+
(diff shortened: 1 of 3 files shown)


nicolastakashi (approved):

matej-g (approved):
Thank you @Vizzyy! This indeed makes sense, good point. Let's fix this 👍.

Vizzyy:
Thanks @matej-g -- I rebased to ensure my commit was signed.

PR #339, opened 2023-12-01
[otel-integration] feat: add support for gke/autopilot

povilasv:
# Description

 Fixes https://coralogix.atlassian.net/browse/ES-152

# How Has This Been Tested?

Run in gke autopilot cluster

# Checklist:

### DIFF
--- a/otel-integration/k8s-helm/gke-autopilot-values.yaml
+++ b/otel-integration/k8s-helm/gke-autopilot-values.yaml
@@ -13,109 +13,19 @@ global:
 distribution: ""
 opentelemetry-agent:
   enabled: true
-  mode: daemonset
   distribution: "gke/autopilot"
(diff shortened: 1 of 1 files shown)


otel-integration/k8s-helm/gke-autopilot-values.yaml
```diff
@@ -0,0 +1,503 @@
+global:
```
matej-g, line 1: Optional: I wonder if we could limit the parameters only those that are extra / changed in comparison to standard `values.yaml`, similar to how we do it for CRD and Windows "special" values file, since the values get merged during installation with defaults.

That would limit the scope and need for changes when we adjust main `values.yaml`, i.e. we would prevent having to duplicate the configuration changes in both `values.yaml` and `gke-autopilot-values.yaml`.

PR #361, opened 2024-02-05
feat: add tail sampling

povilasv:
# Description

Fixes ES-199

This allows users to use [tail-sampling-values.yaml](https://github.com/coralogix/telemetry-shippers/compare/tail-sampling?expand=1#diff-f53c564ac638e8fabccf5447e6b17ab8369ca8e38059f541b791b3c3448b285e) to quickly spin up additional opentemeletry-collector-gateway deployment.

Agents will send data to the gateway, while gateway will do tail sampling descisions

# How Has This Been Tested?

- kind cluster locally

# Checklist:


otel-integration/k8s-helm/tail-sampling-values.yaml
```diff
@@ -0,0 +1,74 @@
...
+          [
+            {
+              name: errors-policy,
+              type: status_code,
+              status_code: {status_codes: [ERROR]}
+            },
+            {
+              name: randomized-policy,
+              type: probabilistic,
+              probabilistic: {sampling_percentage: 25}
```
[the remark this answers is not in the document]
royfur: it is just example for the override file, but we can change the example to 10% if it is more common use case

otel-integration/k8s-helm/values.yaml
```diff
@@ -545,3 +545,90 @@ opentelemetry-cluster-collector:
 
 opentelemetry-agent-windows:
   enabled: false
+opentelemetry-gateway:
+  enabled: false
```
matej-g, line 549: Is there a reason to include the values here instead of in the override file, if we're disabling the sub-chart in the default values file?

otel-integration/k8s-helm/README.md
```diff
@@ -142,6 +142,19 @@ helm upgrade --install otel-coralogix-integration coralogix-charts-virtual/otel-
...
 ```
 
+### Enabling Tail Sampling
+
+If you want to use [Tail Sampling](https://opentelemetry.io/docs/concepts/sampling/#tail-sampling) to reduce the amount of traces using [tail sampling processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/tailsamplingprocessor) you can install `otel-integration` using `tail-sampling-values.yaml` values. For example:
+
+```bash
+helm repo add coralogix-charts-virtual https://cgx.jfrog.io/artifactory/coralogix-charts-virtual
+
+helm upgrade --install otel-coralogix-integration coralogix-charts-virtual/otel-integration \
```
matej-g, line 152: For this use case, should users install only with `tail-sampling-values.yaml` _or_ both default `values.yaml` and `tail-sampling-values.yaml` together like with our other "special" value files?

PR #366, opened 2024-02-07
Fix trace pipeline warning during install

matej-g:
# Description

Fixes https://coralogix.atlassian.net/browse/ES-203.

Fix warning that shows up when installing the chart - this is due to mismatch between the value types (null vs object). Simply removing the tracing pipeline from config here, since it's not needed (upstream chart will still add a dummy trace pipeline that exports to debug).

# How Has This Been Tested?

Locally with `kind`.

# Checklist:

### DIFF
--- a/otel-integration/CHANGELOG.md
+++ b/otel-integration/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## OpenTelemtry-Integration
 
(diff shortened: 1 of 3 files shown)


matej-g:
The docs CI failure is strange, when manually accessing the link it works fine

PR #372, opened 2024-02-19
fix the issue with check action regarding link to issues labels

guyrenny:
# Description

The check action is failing on links for ps labels, so i have added a regex in .mdox.validate.yaml to ignore them.
 Fixes #

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/.mdox.validate.yaml
+++ b/.mdox.validate.yaml
@@ -4,6 +4,9 @@ cache:
   type: 'SQLite'
 
 validators: 
+  # Ignore links to pr lables.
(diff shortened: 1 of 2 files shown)


matej-g:
I'm rather wondering why it is failing, since those links work correctly when manually opened in a browser 🤔

guyrenny:
> I'm rather wondering why it is failing, since those links work correctly when manually opened in a browser 🤔

honestly, I am not sure, for some reason the action started to fail. There was no change to the links or the action code so i am not sure what happened

PR #376, opened 2024-03-01
[otel-integration] add otel-integration version header

povilasv:
# Description

Adding a new global field that will have otel-integration version and then passes it via header.  We will need to maintain Chart.yaml with this global value.

In helm it looks like it is not possible to pass the  parent chart version. {{Chart.Version}} always renders subchart version. I tried to hack around it with tpl and other ways but didnt find a solution.

Also added a CI job to verify that this value matches the Chart.yaml. Example CI failure -> https://github.com/coralogix/telemetry-shippers/actions/runs/8108799144/job/22162697146?pr=376

Fixes ES-212

# How Has This Been Tested?

- kind cluster

# Checklist:

### DIFF
--- a/.github/workflows/otel-integration-version-check.yml
+++ b/.github/workflows/otel-integration-version-check.yml
@@ -1,0 +1,26 @@
+name: Otel Integration Version Check
+
+on:
+  pull_request:
+    paths:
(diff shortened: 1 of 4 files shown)


matej-g (approved):
Nicely done with the workaround 👍

PR #382, opened 2024-03-07
[otel-integration] Bump collector version to 0.96.0 and use upstream target allocator

matej-g:
# Description

Fixes https://coralogix.atlassian.net/browse/ES-215 and https://coralogix.atlassian.net/browse/ES-189

- Bump collector to 0.96.0, including the Windows image
- Use upstream target allocator image, removing the need for us to have our custom fork 🥳 

# How Has This Been Tested?

Locally with `kind`

# Checklist:

### DIFF
--- a/otel-integration/k8s-helm/values.yaml
+++ b/otel-integration/k8s-helm/values.yaml
@@ -5,7 +5,7 @@ global:
   defaultSubsystemName: "integration"
   logLevel: "warn"
   collectionInterval: "30s"
-  version: "0.0.61"
+  version: "0.0.62"
 
   extensions:
     kubernetesDashboard:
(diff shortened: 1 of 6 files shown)


otel-integration/k8s-helm/values.yaml
```diff
@@ -55,11 +55,8 @@ opentelemetry-agent:
     prometheusCR:
       enabled: true
     image:
-      # Use custom image for target allocator until
-      # the official image is updated with the "per node" strategy
-      # https://github.com/open-telemetry/opentelemetry-operator/pull/2430
-      repository: coralogixrepo/target-allocator
-      tag: sha-abe05c5
+      repository: ghcr.io/open-telemetry/opentelemetry-operator/target-allocator
+      tag: v0.95.0
```
matej-g, line 59: Just FYI, operator is yet to have `0.96.0` release, so I'm using `0.95.0`, but the version are compatible just fine.

PR #412, opened 2024-05-28
feat: upgrading chart dependencies

nicolastakashi:
# Description

Upgrading Helm Chart Dependencies.

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/otel-integration/CHANGELOG.md
+++ b/otel-integration/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## OpenTelemtry-Integration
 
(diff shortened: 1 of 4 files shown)


matej-g (commented):
Looks good, but could we add an entry into the `values.yaml`, even if commented out? It would good to somehow (self)document if we're adding this feature, so that users know, as right now you would need to dive into Helm chart code to notice it. 

As well as mention it in the changelog.

matej-g (approved):
chefs kiss

PR #429, opened 2024-07-18
[otel-integration] Add hostmetrics process preset

povilasv:
# Description

Fixes ES-290

# How Has This Been Tested?

kind cluster

# Checklist:

### DIFF
--- a/otel-integration/CHANGELOG.md
+++ b/otel-integration/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## OpenTelemtry-Integration
 
+### v0.0.87 / 2024-07-18
+- [FEAT] Add process option to hostmetrics preset to scrape process metrics.
+
 ### v0.0.86 / 2024-07-08
 - [FEAT] Update Windows collector image to `0.104.0`
 - [FEAT] Update values file to be in sync with Linux agents.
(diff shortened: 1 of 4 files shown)


otel-integration/k8s-helm/values.yaml
```diff
@@ -121,6 +121,10 @@ opentelemetry-agent:
       enabled: true
     hostMetrics:
       enabled: true
+      # Enables process metrics scraping.
+      # Disabled by default, requires privilleged mode
+      process:
```
matej-g, line 126: Should we add the preset also to Windows and other values files?

matej-g (approved):
Looks good, one optional suggestion

otel-integration/k8s-helm/values.yaml
matej-g: According to docs Windows should be supported in `process` scraper: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/README.md#getting-started

Either way I would suggest either adding it or document it's not supported or something. I'll leave it up to you. The PR looks good otherwise 👍

otel-integration/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## OpenTelemtry-Integration
 
+### v0.0.87 / 2024-07-18
+- [FEAT] Add process option to hostmetrics preset to scrape process metrics.
```
nicolastakashi, line 6: Would be nice some documentation about this preset.
1. Metrics Exposed
2. Expected Cardinality
3. Use Case

I think it would be useful, don't need to be in this PR

PR #466, opened 2024-10-25
[ES-364] e2e add telemetrygen for traces

ryantanjunming:
# Description

 Fixes #

# How Has This Been Tested?

# Checklist:


otel-integration/k8s-helm/e2e-test/main_test.go
ryantanjunming, line 1: how should i be writing the traces portion of this e2e test?

otel-integration/k8s-helm/e2e-test/sinks.go
```diff
@@ -0,0 +1,198 @@
...
+		return count >= entriesNum // Changed > to >=
+	}, time.Duration(timeoutSeconds)*time.Second, 1*time.Second,
+		"failed to receive %d entries in %d seconds",
+		entriesNum, timeoutSeconds)
+}
+
+func ScanTracesForAttributes(t *testing.T, ts *consumertest.TracesSink, expectedService string, kvs map[string]ExpectedValue, scopeSpanAttrs []map[string]ExpectedValue) {
+	for i := 0; i < len(ts.AllTraces()); i++ {
+		traces := ts.AllTraces()[i]
+		for i := 0; i < traces.ResourceSpans().Len(); i++ {
```
douglascamata, line 88: The two loops use the same variable for control. This won't behave well. I see that  `ts.AllTraces()` is an array, so the normal `for-range` loop can be used. The same does not apply for `traces.ResourceSpans()`, so that C-style for loop has to stay for this one.

otel-integration/k8s-helm/e2e-test/sinks.go
```diff
@@ -0,0 +1,198 @@
...
+				assert.NoError(t, err)
+			}
+
+			return
+		}
+	}
+	t.Fatalf("no spans found for service %s", expectedService)
+}
+
+func assertExpectedAttributes(attrs pcommon.Map, kvs map[string]ExpectedValue) error {
```
douglascamata, line 124: I think this whole function could be simplified to something like this:

```go
var notFoundAttrs []string

attrs.Range(func(k string, v pcommon.Value) bool {
	if val, found := kvs[k]; found {
		switch val.Mode {
		case AttributeMatchTypeEqual:
			if val.Value == v.AsString() {
				foundAttrs[k] = true
			}
		case AttributeMatchTypeRegex:
			matched, _ := regexp.MatchString(val.Value, v.AsString())
			if matched {
				foundAttrs[k] = true
			}
		case AttributeMatchTypeExist:
			foundAttrs[k] = true
		}
	} else {
		notFoundAttrs = append(notFoundAttrs, k)
	}
	return true
})

// if something is missing, add a summary with an overview of the expected and actual attributes for easier troubleshooting
if len(notFoundAttrs) > 0 {
    expectedJson, _ := json.MarshalIndent(kvs, "", "  ")
	actualJson, _ := json.MarshalIndent(attrs.AsRaw(), "", "  ")
	err = errors.Join(err, fmt.Errorf("one or more attributes were not found.\nExpected attributes:\n %s \nActual attributes: \n%s", expectedJson, actualJson))
}

return err
```

otel-integration/k8s-helm/e2e-test/sinks.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// ScanForServiceMetrics asserts that the metrics sink provided in the arguments
+// contains the given metrics for a service
+func ScanForServiceMetrics(t *testing.T, ms *consumertest.MetricsSink, expectedService string, expectedMetrics []string) {
+	for _, r := range ms.AllMetrics() {
+		for i := 0; i < r.ResourceMetrics().Len(); i++ {
+			resource := r.ResourceMetrics().At(i).Resource()
+			service, exist := resource.Attributes().Get(ServiceNameAttribute)
+			assert.Equal(t, true, exist, "resource does not have the 'service.name' attribute")
```
douglascamata, line 173: You can use `assert.True(t, exist, msg)`

douglascamata (commented):
Awesome job, @ryantanjunming. I left some comments with suggestions and some are purely informative. I'm aware this is still a draft, so no worries.

otel-integration/k8s-helm/e2e-test/main_test.go
```diff
@@ -4,52 +4,74 @@
...
-	deadline := time.Now().Add(wait)
-	for {
-		if time.Now().After(deadline) {
-			break
+	telemetryGenObjs, telemetryGenObjInfos := k8stest.CreateTelemetryGenObjects(t, k8sClient, createTeleOpts)
+	defer func() {
+		for _, obj := range telemetryGenObjs {
+			require.NoErrorf(t, k8stest.DeleteObject(k8sClient, obj), "failed to delete object %s", obj.GetName())
 		}
+	}()
```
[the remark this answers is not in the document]
ryantanjunming: great idea

otel-integration/k8s-helm/e2e-test/main_test.go
```diff
@@ -4,52 +4,74 @@
...
+	metricsConsumer := new(consumertest.MetricsSink)
+	tracesConsumer := new(consumertest.TracesSink)
+	shutdownSink := StartUpSinks(t, metricsConsumer, tracesConsumer)
+	defer shutdownSink()
+
+	testID := uuid.NewString()[:8]
+	createTeleOpts := &k8stest.TelemetrygenCreateOpts{
+		ManifestsDir: filepath.Join(k8sDir, "telemetrygen"),
+		TestID:       testID,
+		OtlpEndpoint: "", //unused
```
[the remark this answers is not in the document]
ryantanjunming: thanks

otel-integration/k8s-helm/e2e-test/main_test.go
```diff
@@ -4,52 +4,74 @@
...
-	waitTime := 2 * time.Minute
-	waitForData(waitTime)
+	//Check if the HOST_ENDPOINT is set
+	require.Equal(t, k8stest.HostEndpoint(t), os.Getenv("HOSTENDPOINT"), "HostEndpoints does not match env and detected")
 
-	checkResourceMetrics(t, metricsConsumer.AllMetrics())
-}
+	k8sDir := filepath.Join("k8s")
+
+	kubeConfigPath := os.Getenv("KUBECONFIG")
```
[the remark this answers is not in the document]
ryantanjunming: moving to client.go with assert testing

otel-integration/k8s-helm/e2e-test/main_test.go
```diff
@@ -151,3 +174,17 @@ func checkResourceAttributes(t *testing.T, attributes pcommon.Map, scopeName str
 
 	return nil
 }
+
+func checkGeneratedTraces(t *testing.T, actual []ptrace.Traces, testID string) error {
```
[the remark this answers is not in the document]
ryantanjunming: testID is in the k8s manifest name, so will check pod names and deployment names etc

otel-integration/k8s-helm/e2e-test/sinks.go
ryantanjunming: thanks, yeah less loops

otel-integration/k8s-helm/e2e-test/sinks.go
```diff
@@ -0,0 +1,198 @@
...
+	var actualMetrics []string
+	for i := 0; i < sm.Len(); i++ {
+		actualMetrics = append(actualMetrics, sm.At(i).Name())
+	}
+
+	for _, m := range expectedMetrics {
+		if !slices.Contains(actualMetrics, m) {
+			return fmt.Errorf("metric: %s not found", m)
+		}
+	}
```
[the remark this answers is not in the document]
ryantanjunming: removed checking of metrics from telemetrygen (only traces are needed)

otel-integration/k8s-helm/e2e-test/sinks.go
ryantanjunming: oh i'll change it back cause the  `fmt.Errorf("attribute '%v' not found", k)` is quite useful when debugging

otel-integration/k8s-helm/e2e-test/sinks.go
```diff
@@ -0,0 +1,198 @@
...
+	t.Fatalf("no metric found for service %s", expectedService)
+}
+
+func assertExpectedMetrics(expectedMetrics []string, sm pmetric.MetricSlice) error {
+	var actualMetrics []string
+	for i := 0; i < sm.Len(); i++ {
+		actualMetrics = append(actualMetrics, sm.At(i).Name())
+	}
```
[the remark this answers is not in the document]
ryantanjunming: removed metrics testing, so irrelavant for now, thanks

otel-integration/k8s-helm/e2e-test/sinks.go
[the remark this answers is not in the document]
ryantanjunming: removed metrics testing, so irrelevant for now, thanks

otel-integration/k8s-helm/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: otel-integration
 description: OpenTelemetry Integration
-version: 0.0.114
+version: 0.0.115
```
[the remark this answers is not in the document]
ryantanjunming: @royfur my previous e2e changes, @povilasv said it was a requirement for even the small changes

https://github.com/coralogix/telemetry-shippers/pull/461/files

otel-integration/k8s-helm/e2e-test/main_test.go
```diff
@@ -89,7 +112,9 @@ func checkResourceMetrics(t *testing.T, actual []pmetric.Metrics) error {
 	}
 
 	if len(missingMetrics) > 0 {
-		t.Fatalf("metrics %v were not found in the actual metrics", missingMetrics)
+		// Note: actual metrics should a subset of expected metrics,
+		// and some expected metrics may not be found in actual metrics
+		t.Logf("Note: expected metrics %v were not found in the actual metrics", missingMetrics)
```
douglascamata, line 117: Is there a way to make the actual metrics contain all the expected metrics? Otherwise this fact can let some issues go unnoticed.

otel-integration/k8s-helm/e2e-test/main_test.go
```diff
@@ -142,12 +179,88 @@ func checkResourceAttributes(t *testing.T, attributes pcommon.Map, scopeName str
...
+		for i := 0; i < trace.ResourceSpans().Len(); i++ {
+			rspans := trace.ResourceSpans().At(i)
+
+			_, ok := expectedTracesSchemaURL[rspans.SchemaUrl()]
+			require.True(t, ok, "traces: resource %v does not match one of the expected values", rspans.SchemaUrl())
+			if ok {
+				expectedTracesSchemaURL[rspans.SchemaUrl()] = true
+			}
+
+			resource := trace.ResourceSpans().At(i).Resource()
```
douglascamata, line 210: Why not reuse the variable you already have?

```suggestion
			resource := rspans.Resource()
```

otel-integration/k8s-helm/e2e-test/main_test.go
```diff
@@ -142,12 +179,88 @@ func checkResourceAttributes(t *testing.T, attributes pcommon.Map, scopeName str
...
+			service, exist := resource.Attributes().Get(ServiceNameAttribute)
+
+			expectedTrace := expectedTraces(testID, testNs)[service.AsString()]
+			require.NotEmpty(t, expectedTrace, "traces: unexpected service name %v", service.AsString())
+			require.True(t, exist, "traces: resource does not have the 'service.name' attribute")
+			assert.NoError(t, assertExpectedAttributes(resource.Attributes(), expectedTrace.attrs))
+
+			require.NotZero(t, trace.ResourceSpans().At(i).ScopeSpans().Len())
+			require.NotZero(t, trace.ResourceSpans().At(i).ScopeSpans().At(0).Spans().Len())
```
douglascamata, line 219: Could use reuse `rspans` here.

```suggestion
			require.NotZero(t, rspans.At(i).ScopeSpans().Len())
			require.NotZero(t, rspans.At(i).ScopeSpans().At(0).Spans().Len())
```

otel-integration/k8s-helm/e2e-test/sinks.go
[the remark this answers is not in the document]
douglascamata: This stays true, no?

otel-integration/k8s-helm/e2e-test/sinks.go
```diff
@@ -0,0 +1,198 @@
...
+		entriesNum, timeoutSeconds)
+}
+
+func WaitForTraces(t *testing.T, entriesNum int, tc *consumertest.TracesSink) {
+	timeoutSeconds := 180 // 3 minutes
+	require.Eventuallyf(t, func() bool {
+		count := len(tc.AllTraces())
+		t.Logf("Waiting for traces: got %d/%d", count, entriesNum)
+		return count >= entriesNum // Changed > to >=
+	}, time.Duration(timeoutSeconds)*time.Second, 1*time.Second,
```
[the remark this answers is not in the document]
douglascamata: This stays unresolved.

otel-integration/k8s-helm/e2e-test/testdata/values-e2e-test.yaml
```diff
@@ -11,10 +32,9 @@ opentelemetry-agent:
   config:
     exporters:
       otlp:
-        endpoint: 172.17.0.1:4317 # Use the default Docker host gateway IP
+        endpoint: ${env:HOSTENDPOINT}:4317
```
douglascamata, line 35: I would advice against relying on shell env var expansion here. If the OTEL exporter code changes and somehow starts to validate this endpoint address before env var expansion happens, we will have issues. I've fixed a similar bug recently in our OTEL collector chart.

I think we can rely on `host.docker.internal`, which has been adopted by a few tools (podman, colima, etc) already as the de-facto standard hostname for this.

douglascamata (changes requested):
This is looking good!

I have a few new comments and some old ones that are still not resolved.

otel-integration/k8s-helm/e2e-test/main_test.go
ryantanjunming: the set of metrics changes slightly since I've added the run of telemetrygen pods and traces are coming in. One example are the pod deletion which can occur before or after the sink ends so the pod deletion metric may or may not appear.

otel-integration/k8s-helm/e2e-test/testdata/values-e2e-test.yaml
ryantanjunming: changes made to insert via helm config param, works

otel-integration/k8s-helm/e2e-test/main_test.go
ryantanjunming: made the sink longer (5mins) and metrics seem more consistent now.

otel-integration/k8s-helm/e2e-test/sinks.go
ryantanjunming: removed this function, so would be irrelevant

otel-integration/k8s-helm/e2e-test/sinks.go
ryantanjunming: still needs a second loop to print out the unfounded attributes

otel-integration/k8s-helm/e2e-test/sinks.go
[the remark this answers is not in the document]
ryantanjunming: fixed

ryantanjunming:
Will reference the oteltest as a package in the next e2e ticket, thanks

PR #474, opened 2024-11-26
FEAT - ECS Fargate Resource Catalog support

MichaelBriggs-Coralogix:
# Description

Adding Resource Catalog Support to ECS Fargate integration

# How Has This Been Tested?

Deployed parameter store manually and with cloudformation template. Attached new config to existing ECS Fargate deployment. Observed Logs, Metrics and Traces from example workload applications.

# Checklist:


otel-ecs-fargate/config.yaml
```diff
@@ -1,127 +1,134 @@
-exporters:
```
MichaelBriggs-Coralogix, line 1: In case it's not clear, the exporter still exists, it's just that I put receivers first, then processors and then exporters. Just rearranged things.

royfur (approved):

PR #549, opened 2025-04-11
[otel-integration] docs: for windows opamp issue

povilasv:
# Description

Fixes ES-555

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/otel-integration/k8s-helm/README.md
+++ b/otel-integration/k8s-helm/README.md
@@ -1778,6 +1778,16 @@ presets:
     enabled: true
 ```
 
+#### Known errors
+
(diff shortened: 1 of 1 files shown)


otel-integration/k8s-helm/README.md
```diff
@@ -1778,6 +1778,16 @@ presets:
     enabled: true
 ```
 
+#### Known errors
+
+When running on Windows, you might see the "failed getting host info" error. This is expected behavior because the collector attempts to retrieve Windows metadata from the Windows Registry, which is only possible when running from HostProcess Windows containers.
```
douglascamata, line 1783: I would add to this that it has no negative impact in functionality of the Collector or OpAMP in any way.

PR #571, opened 2025-05-06
update/CDS-2086: rewrite fluent-bit k8s configuration to YAML

cazorla19:
# Description

Fixes CDS-2086.

> Fluent Bit traditionally offered a classic configuration mode, a custom configuration format that we are gradually phasing out. While classic mode has served well for many years, it has several limitations. Its basic design only supports grouping sections with key-value pairs and lacks the ability to handle sub-sections or complex data structures like lists.
> 
> YAML, now a mainstream configuration format, has become essential in a cloud ecosystem where everything is configured this way. To minimize friction and provide a more intuitive experience for creating data pipelines, we strongly encourage users to transition to YAML. The YAML format enables features, such as processors, that are not possible to configure in classic mode.
> 
> As of Fluent Bit v3.2, you can configure everything in YAML.

This will move fluent-bit configuration to YAML in order to avoid doing it too late, when classic mode will be deprecated.

# Checklist

* `k8s-manifests`: migrated
* `k8s-helm-http`: migrated
* `k8s-helm-coralogix`: depends on older version (0.0.3)
* `ecs-fargate`: depends on https://github.com/aws/aws-for-fluent-bit , which is dependent on older version (1.9.10)
* `eks-fargate`: depends on https://github.com/aws/aws-for-fluent-bit , which is dependent on older version (1.9.10)

# How Has This Been Tested?

Run both manifests and helm chart on local k8s cluster, compared output to Coralogix with previous version to make sure there is no difference.

# Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ Chart.lock
 # Fluent-bit
 logs/fluent-bit/test/
 logs/fluent-bit/k8s-manifest/fluentbit-env-cm.yaml
+**/charts/
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@
 .idea
 .vscode
 Chart.lock
+
+# Fluent-bit
+logs/fluent-bit/test/
+logs/fluent-bit/k8s-manifest/fluentbit-env-cm.yaml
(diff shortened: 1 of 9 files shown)


juan-coralogix (approved):
LGTM

PR #573, opened 2025-05-08
Add E2E test for opentelemetry operator

iblancasa:
# Description

- Add E2E for OpenTelemetry Operator
- Check the `otelcol_otelsvc_k8s_pod_deleted_ratio` metric


.github/workflows/otel-integration-e2e-test.yml
```diff
@@ -54,6 +54,7 @@ jobs:
           KUBECONFIG: ${{ env.KUBECONFIG }}
         run: |
           cd ./otel-integration/k8s-helm/e2e-test/
+          go clean -testcache
```
douglascamata, line 57: Why this?

douglascamata (approved):

.github/workflows/otel-integration-e2e-test.yml
iblancasa: Since it is a go test, it can happen that the result is being cached. It happened to me while working on this PR... I assume there should not be a problem for GitHub workflows but... I prefer to run this here to just double check cache doesn't play with our feelings.

douglascamata (approved):

PR #575, opened 2025-05-14
[otel-eks-fargate] Fix incorrect overwrite of k8s.node.name

wnjuguna:
# Description

The value of `k8s_node_name` label set in telemetry from this integration is incorrect. It is incorrectly set to the **node** where the `fargate-otel-collector pod` is running

Fixes # CDS-2130

# How Has This Been Tested?
Verified in lab and with customer that the correct value is set after removing the k8s.node.name upsert in the resource/metadata processor.  

# Checklist:


daidokoro (approved):
LGTM

PR #631, opened 2025-07-28
[otel-integration] Fix missing version bump

douglascamata:
# Description

Fixes missing version bump after #630

### DIFF
--- a/otel-integration/k8s-helm/Chart.yaml
+++ b/otel-integration/k8s-helm/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: otel-integration
 description: OpenTelemetry Integration
-version: 0.0.203
+version: 0.0.204
 keywords:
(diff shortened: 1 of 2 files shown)


otel-integration/k8s-helm/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: otel-integration
 description: OpenTelemetry Integration
-version: 0.0.203
+version: 0.0.204
```
nicolastakashi, line 4: Should maybe add a lint to always expect a version bump unless you explicitly add a label to skip-version-bump

douglascamata: Good idea, @nicolastakashi. By chance, have you seen something like this in any other repository that we could use as a base for our own solution?

royfur (approved):

PR #657, opened 2025-09-01
ES-56: port support for EKS Fargate

iblancasa:
# Description

Fixes ES-56.

# How Has This Been Tested?

### DIFF
--- a/otel-integration/CHANGELOG.md
+++ b/otel-integration/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## OpenTelemetry-Integration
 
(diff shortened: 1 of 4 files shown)


douglascamata (changes requested):
Approved by accident.

NimrodAvni78:
@douglascamata actually no since its not a change in the profiler agent, its just a change in the collector, but for the profiler signal
as for changelog you are right, we can add something like 
```
- [Feat] Profiling: improve service name detection by otel conventions.
```

douglascamata (approved):

PR #735, opened 2025-12-24
[CX-4142]: Added script for sync EC2 integration collector version

lzankivl:
https://coralogix.atlassian.net/browse/CX-25474


v-tec2706 (commented):
LGTM but please test it before merging

PR #737, opened 2025-12-24
Add E2E tests for span metrics presets to ensure we don't break backward compatibility

iblancasa:
# Description
CX-13396

### DIFF
--- a/otel-integration/k8s-helm/e2e-test/testdata/values-e2e-span-metrics.yaml
+++ b/otel-integration/k8s-helm/e2e-test/testdata/values-e2e-span-metrics.yaml
@@ -1,6 +1,10 @@
 # Values file for span metrics E2E tests
 # Uses default preset configuration to detect any breaking changes in the preset itself
 
+global:
+  logLevel: "info"
+  hostedEndpoint: "host.docker.internal"
+
 opentelemetry-agent:
   presets:
     spanMetrics:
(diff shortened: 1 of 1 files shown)


.github/workflows/otel-integration-e2e-test.yml
```diff
@@ -130,6 +130,66 @@ jobs:
...
 
+  otel-integration-span-metrics-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Install Go
+        uses: actions/setup-go@v3
```
douglascamata: Latest version of this is v6, could we use it?

.github/workflows/otel-integration-e2e-test.yml
```diff
@@ -130,6 +130,66 @@ jobs:
...
+        with:
+          go-version: 1.21.x
```
douglascamata: Maybe we should upgrade this a little bit? 1.21 is quite old.

douglascamata (approved):
LGTM, just a small note regarding using some very old versions of certain things.

.github/workflows/otel-integration-e2e-test.yml
iblancasa: Yes, you are right.

PR #868, opened 2026-04-10
fix(CDS-2869): add IsMap(body) guard to transform/kube-events processor

natnayr:
## Problem

The `transform/kube-events` processor in the otel-infrastructure-collector Helm chart uses an OTTL `keep_keys` statement to filter Kubernetes event log fields:

```yaml
- keep_keys(body, ["type", "action", "eventTime", ...])
```

This fails with `INVALID_ARGUMENT` when a k8s event log body is a **plain string** rather than a JSON object. Kubernetes can emit string-body events (e.g., from certain controllers or when the event message itself is not structured), causing the entire batch to fail and drop.

## Fix

Add an inline `where IsMap(body)` guard so the transformation only runs when the body is actually a map:

```yaml
- keep_keys(body, ["type", "action", "eventTime", ...]) where IsMap(body)
```

String-bodied records pass through untouched. Map-bodied records are filtered as before.

**Why inline `where` and not the `conditions:` block?**  
The `conditions:` block syntax was introduced after OTel Collector v0.85.0. Using `where IsMap(body)` inline is compatible with v0.85.0 and all later versions.

## Testing

Locally reproduced with a 3-record OTLP test:
- Record 1: string body → previously caused INVALID_ARGUMENT, now passes through unchanged
- Record 2: map body with an extra field → extra field correctly removed by keep_keys
- Record 3: map body with valid keys only → all keys preserved

Both `values.yaml` and `ci/ci-values.yaml` updated.

Jira: [CDS-2869](https://coralogix.atlassian.net/browse/CDS-2869)

[CDS-2869]: https://coralogix.atlassian.net/browse/CDS-2869?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/fluentd-ecs-dockerhub-description.yml
+++ b/.github/workflows/fluentd-ecs-dockerhub-description.yml
@@ -10,6 +10,8 @@ on:
       - 'logs/fluentd/aws-ecs/readme.md'
   workflow_dispatch:
 
+    permissions:
+      contents: read
 jobs:
   update-description:
     runs-on: ubuntu-latest
--- a/.github/workflows/fluentd-ecs-dockerhub-description.yml
+++ b/.github/workflows/fluentd-ecs-dockerhub-description.yml
@@ -1,0 +1,23 @@
+name: Fluentd-ECS DockerHub Description
+
+on:
+  push:
+    branches: master
+    paths:
+      - 'logs/fluentd/aws-ecs/readme.md'
+  workflow_dispatch:
+
+jobs:
+  update-description:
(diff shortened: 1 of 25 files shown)


natnayr:
Wrong directory — fix should be in otel-integration/k8s-helm/, not otel-infrastructure-collector/k8s-helm/. Reopening with correct location.

PR #871, opened 2026-04-13
feat(e2e): add verify Go module, CLI, and patch scripts

dmenilo1:
## Summary

  Adds the foundation for a new unified E2E framework.
  The framework will eventually cover all required deployment targets (K8S, ECS, Linux, Windows...) using a single shared verification module, per-target
  expectation files, and per-target jobs in a single GitHub Actions workflow.

  ## What this PR adds

  ```
  e2e/
  ├── go.mod / go.sum                       # Go module: coralogix.com/telemetry-shippers/e2e
  ├── verify/
  │   ├── coralogix.go                      # DataPrime + PromQL HTTP clients (with parsed-row access)
  │   ├── coralogix_test.go
  │   ├── verify.go                         # Verify{Logs,Traces,Metrics} with backoff polling
  │   ├── verify_test.go
  │   ├── expectations.go                   # YAML schema + loader for per-target expectations
  │   ├── expectations_verify.go            # VerifyExpectations() runs structured assertions
  │   └── expectations_test.go
  ├── cmd/e2e-verify/
  │   └── main.go                           # CLI wrapping the verify module
  ├── scripts/
  │   ├── patch-config-runid.sh             # Inject e2e.run_id (Linux/macOS)
  │   ├── patch-config-runid.ps1            # Same for Windows
  │   └── wait-healthy.sh                   # Poll collector :13133 until ready
  └── expectations/
      ├── k8s.yaml                          # Per-target assertions
      ├── ecs-ec2.yaml
      ├── linux-standalone.yaml
      ├── windows-standalone.yaml
      └── macos-standalone.yaml
  ```

  ## How it works

  For each E2E test run:

  1. The orchestrator deploys the OTel collector with the chart's `build/otel-config.yaml`, after `patch-config-runid.sh` injects a unique `e2e.run_id` resource
   attribute.
  2. `wait-healthy.sh` polls `:13133/healthcheck` until the collector is ready.
  3. `e2e-verify` queries Coralogix DataPrime + PromQL APIs to confirm telemetry with the matching `e2e.run_id` arrived. Two modes:
     - `--check logs --check traces --check metrics` — existence-only (debugging).
     - `--expectations <file>` — structured per-target checks: `min_count`, `required_attributes` (resource attrs on logs/traces), `required_labels` (PromQL
  labels on metrics), and `collector_health.max_send_failures` (catches collector-side ingest failures via `otelcol_exporter_send_failed_records_total`).
  4. Cleanup (uninstall / `terraform destroy`) runs in `if: always()`.

  The verify CLI is signal-agnostic and works for all future targets without modification — only the per-target `expectations/<target>.yaml` differs.

  ## Expectations schema

  Each per-target file describes what the chart **should** emit. Sections are optional; missing sections aren't checked.

  ```yaml
  logs:
    min_count: 1
    required_attributes:
      - host.name
      - service.name
      - e2e.run_id

  traces:
    min_count: 1
    required_attributes:
      - service.name
      - e2e.run_id

  metrics:
    min_count: 1
    required_labels:
      - cx_application_name
      - host_name
      - e2e_run_id

  collector_health:
    max_send_failures: 0
  ```

  ## Validated end-to-end

  Two manual runs against real Coralogix using `otel-macos-standalone`:

  **Run 1 — existence-only mode (`--check`):**
  ```
  ✅ logs:    found=true, count=5
  ✅ metrics: found=true, count=9137
     OK: all checks passed
  ```

  **Run 2 — expectations mode (`--expectations expectations/macos-standalone.yaml`):**
  ```
  ✅ logs (4 attrs):    100 rows, all attrs present on every row
  ✅ metrics (4 labels): 9265 series, all labels present on every series
  ✅ collector_health:  no send failures
     OK: all expectations met
  ```

  Failure path tested by deliberately requiring an attribute the chart strips (`cx.otel_integration.name`) — produced clear, structured output pinpointing the
  failed assertion.

  ## What's NOT in this PR (follow-up phases)

  - Per-target jobs (K8S, ECS, Linux, Windows, macOS) — separate commits on the same branch
  - ECS bootstrap Terraform (persistent infra) — separate commit
  - Orchestrator workflow (`e2e-otel-release-test.yml`) — added when first job is wired up
  - Nightly cleanup workflow
  - Migration / deprecation of the old ECS test in `aws-lambda-integration-tests/ecs_ec2_test`

  ## Test plan

  - [x] Unit tests pass: `cd e2e && go test ./...` (21 tests: 12 client/verify + 9 expectations)
  - [x] `go vet ./...` clean
  - [x] CLI builds and shows usage
  - [x] `patch-config-runid.sh` tested on all three standalone build configs (linux/windows/macos) — correctly targets `processors."resource/metadata"`,
  preserves existing attributes, idempotent, fails loudly if processor not in any pipeline
  - [x] End-to-end against real Coralogix in **both** modes (existence + expectations) on macOS standalone
  - [x] Failure path tested — required attribute the chart strips → clear structured failure
  - [x] `wait-healthy.sh` timeout path tested

  ## Notes for reviewers

  - **Two API keys are required.** `E2E_CX_API_KEY` (Send-Your-Data) for the collector to ingest into CX. `E2E_CX_LOGS_QUERY_KEY` (Logs Query Key) for the
  verify CLI to query CX. The Send-Your-Data key alone cannot query.
  - DataPrime auth header is the raw key (no `Bearer` prefix). PromQL auth header uses `Bearer <key>`. Both confirmed against real CX.
  - Default `--window` is 24h to handle cases where the collector picks up pre-existing log files with old timestamps.
  - `InitialDelay` defaults to 0 in the Go package; the CLI sets it to 30s explicitly to give the collector time to flush its first batch.
  - **DataPrime path-syntax quirk:** structured paths only support `[a-zA-Z0-9_]`, so OTel attribute keys with dots (e.g. `e2e.run_id`) cannot be queried via
  `$d.resource.attributes.X` — this returns `keypath does not exist`. The default logs/traces filter uses DataPrime's `~~` substring operator on `$d`, which is
  reliable since run_ids are unique. Filters are overridable via `--logs-filter`, `--traces-filter`, `--metrics-filter` flags using `{{run_id}}` as the
  placeholder.
  - **`collector_health` limitation:** `otelcol_exporter_send_failed_records_total` only increments for collector-side failures (network, auth, dropped queue).
  Coralogix-side rejection of oversized labels (e.g. `max_label_length` fair-usage breach) does **not** increment this metric — the collector successfully sends
   the data, CX silently drops the offending labels. A separate check that queries CX warning logs would be needed for full coverage; out of scope for this PR.

  ## Side finding (separate ticket worth filing)

  The `otel-macos-standalone` chart emits per-process metrics with `process.command_line` labels that exceed Coralogix's 1024-char limit on desktops with many
  GUI apps (Chrome, Slack, Electron). During validation this caused 1,212 `max_label_length` fair-usage breaches. Suggested fix: add a transform processor to
  truncate `process.command_line` to ~256 chars in the chart, or restrict the process scraper. Worth filing for chart owners — exactly the kind of regression
  release E2E should surface.

### DIFF
--- a/e2e/cmd/e2e-verify/main.go
+++ b/e2e/cmd/e2e-verify/main.go
@@ -45,6 +45,7 @@ func main() {
 		logsFilter    = flag.String("logs-filter", "", "Override DataPrime filter clause for logs (use {{run_id}} placeholder)")
 		tracesFilter  = flag.String("traces-filter", "", "Override DataPrime filter clause for traces (use {{run_id}} placeholder)")
 		metricsFilter = flag.String("metrics-filter", "", "Override PromQL selector for metrics (use {{run_id}} placeholder)")
+		expectations  = flag.String("expectations", "", "Path to expectations YAML. When set, runs structured expectation checks instead of --check flags.")
 	)
 	var checks stringSlice
 	flag.Var(&checks, "check", "Signal to verify: logs, traces, or metrics. Repeat for multiple. [required]")
@@ -75,8 +76,8 @@ func main() {
 	if *runID == "" {
 		missing = append(missing, "--run-id")
 	}
-	if len(checks) == 0 {
-		missing = append(missing, "--check (one of: logs, traces, metrics)")
+	if len(checks) == 0 && *expectations == "" {
+		missing = append(missing, "--check (one of: logs, traces, metrics) OR --expectations <file>")
 	}
 	if len(missing) > 0 {
 		fmt.Fprintf(os.Stderr, "ERROR: missing required flags: %s\n\n", strings.Join(missing, ", "))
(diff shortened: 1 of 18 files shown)


juan-coralogix (approved):
👍

PR #938, opened 2026-06-12
Implemented linters in CI

markshleifer-coralogix:
# Description

**Most of the added lines (around 97%) are Helm golden render snapshots.**

Add CI coverage for security and hygiene checks:

- gitleaks current-tree secret scanning
- shellcheck for checked-in shell scripts
- hadolint for Dockerfiles
- Helm golden render checks for high-risk presets

Some `shellcheck`/`hadolint` ignores were added where changing the code would be riskier than documenting the existing baseline. Examples include preserving existing Docker label keys that may be externally consumed, keeping established package-install commands unchanged, and avoiding behavior changes in installer/runtime scripts where lint-only rewrites could affect execution.

# How Has This Been Tested?

1. Deliberately added or changed files to force each new CI job to fail:
   - fake private key for `gitleaks`
   - unquoted shell variable for `shellcheck`
   - bad Dockerfile fixture for `hadolint`
   - golden render mismatch for Helm

2. Confirmed the jobs failed for the expected reasons, then reverted the failure fixtures.

3. Re-ran the checks and verified the scripts/jobs passed again.

4. Built the changed Linux Dockerfiles with Docker where possible; the buildable images passed. Windows Dockerfiles were not built locally because macOS Docker Desktop cannot build Windows container images.

# Checklist:

### DIFF
--- a/.github/workflows/security-hygiene.yml
+++ b/.github/workflows/security-hygiene.yml
@@ -1,0 +1,93 @@
+name: Security and Hygiene
+
+on:
+  pull_request:
+
+permissions:
+  contents: read
+
+env:
+  GITLEAKS_VERSION: 8.24.3
+  SHELLCHECK_VERSION: 0.10.0
+  HADOLINT_VERSION: 2.12.0
+
+jobs:
+  gitleaks:
+    runs-on: ubuntu-latest
+    name: gitleaks PR secret scan
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install gitleaks
+        run: |
+          set -euo pipefail
+          mkdir -p "${RUNNER_TEMP}/bin"
+          curl -sSfL \
+            -o "${RUNNER_TEMP}/gitleaks.tar.gz" \
+            "https://github.com/gitleaks/gitleaks/releases/download/v${GITLEAKS_VERSION}/gitleaks_${GITLEAKS_VERSION}_linux_x64.tar.gz"
+          tar -xzf "${RUNNER_TEMP}/gitleaks.tar.gz" -C "${RUNNER_TEMP}/bin" gitleaks
+          echo "${RUNNER_TEMP}/bin" >> "${GITHUB_PATH}"
+
+      - name: Scan current tree
+        run: gitleaks detect --no-git --source . --redact --verbose
+
+  shellcheck:
+    runs-on: ubuntu-latest
+    name: shellcheck checked-in scripts
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install shellcheck
+        run: |
+          set -euo pipefail
+          mkdir -p "${RUNNER_TEMP}/bin"
+          curl -sSfL \
+            -o "${RUNNER_TEMP}/shellcheck.tar.xz" \
+            "https://github.com/koalaman/shellcheck/releases/download/v${SHELLCHECK_VERSION}/shellcheck-v${SHELLCHECK_VERSION}.linux.x86_64.tar.xz"
+          tar -xJf "${RUNNER_TEMP}/shellcheck.tar.xz" \
+            -C "${RUNNER_TEMP}/bin" \
+            --strip-components=1 \
+            "shellcheck-v${SHELLCHECK_VERSION}/shellcheck"
+          echo "${RUNNER_TEMP}/bin" >> "${GITHUB_PATH}"
+
+      - name: Lint checked-in shell scripts
+        run: |
+          set -euo pipefail
+          mapfile -d '' scripts < <(find . -type f -name '*.sh' -not -path './tmp/*' -print0)
+          shellcheck "${scripts[@]}"
+
+  hadolint:
+    runs-on: ubuntu-latest
+    name: hadolint Dockerfiles
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install hadolint
+        run: |
+          set -euo pipefail
+          mkdir -p "${RUNNER_TEMP}/bin"
+          curl -sSfL \
+            -o "${RUNNER_TEMP}/bin/hadolint" \
+            "https://github.com/hadolint/hadolint/releases/download/v${HADOLINT_VERSION}/hadolint-Linux-x86_64"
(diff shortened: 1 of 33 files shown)


telemetrygen-windows-image/Dockerfile
```diff
@@ -5,16 +5,19 @@ ARG TELEMETRYGEN_VERSION=v0.147.0
 ARG WIN_BASE_IMAGE=mcr.microsoft.com/windows/servercore:ltsc2019
 
 # Build stage: cross-compile telemetrygen for Windows
+# hadolint ignore=DL3029
```
iblancasa, line 9: Can you add what we are ignoring in the different `hadolint ignore=<X>`? Like `Do not use --platform= with FROM`

scripts/version_bump_check.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+# shellcheck disable=SC2086,SC2181,SC2046,SC2005
```
iblancasa: Why ignoring `SC2005` instead of using `printf`?
Why are `SC2181`, `SC2086` and `SC2046` ignored?

otel-supervised-collector/uninstall.sh
```diff
@@ -34,6 +34,7 @@ ensure_sudo() {
 
 detect_pkg_type() {
   if [ -r /etc/os-release ]; then
+    # shellcheck disable=SC1091
```
iblancasa: Why ignoring?

otel-integration/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## OpenTelemetry-Integration
 
+### v0.0.318 / 2026-06-12
+
+- [Chore] Add Helm golden render coverage for high-risk presets and document intentional shellcheck suppressions in the E2E runner.
```
iblancasa: I don't think we need to add a changelog entry for this. This should not be something users will notice a behavior change, right?

otel-integration/k8s-helm/values.yaml
```diff
@@ -5,7 +5,7 @@ global:
   defaultSubsystemName: "integration"
   logLevel: "info"
   collectionInterval: "30s"
-  version: "0.0.317"
```
iblancasa, line 8: No need to release anything new, I guess.

otel-integration/k8s-helm/tests/golden/ebpf-profiler.yaml
```diff
@@ -0,0 +1,2277 @@
+---
```
iblancasa, line 1: How is this file generated?

.github/workflows/security-hygiene.yml
```diff
@@ -0,0 +1,93 @@
...
+  HADOLINT_VERSION: 2.12.0
+
+jobs:
+  gitleaks:
+    runs-on: ubuntu-latest
+    name: gitleaks PR secret scan
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install gitleaks
```
iblancasa, line 21: How about adding some script we can run locally to install these tools in our machines and reuse here? 
Or maybe use one of the GitHub Actions for Hadolint or Gitleaks.

otel-supervised-collector/uninstall.sh
markshleifer-coralogix: SC1091 is ignored because /etc/os-release exists on the target Linux host at runtime, but shellcheck cannot follow that system file during static analysis.

logs/fluent-bit/image/Dockerfile
```diff
@@ -1,6 +1,10 @@
 FROM fluent/fluent-bit:3.2.10
 ARG TARGETARCH
+# DL3048: Invalid label key.
+# hadolint ignore=DL3048
```
povilasv, line 4: Why is this invalid? Maybe we should remove the label then?

otel-supervised-ebpf-profiler/Dockerfile
```diff
@@ -3,11 +3,20 @@ ARG COLLECTOR_VERSION=0.147.0
 ARG SUPERVISOR_VERSION=0.147.0
 
 FROM alpine:3.22 AS certs
+# DL3018: Pin versions in apk add.
+# DL3019: Use --no-cache instead of --update.
+# hadolint ignore=DL3018,DL3019
```
povilasv: can we do this?

logs/fluent-bit/image/Dockerfile
markshleifer-coralogix: Hadolint expects label keys to follow modern lowercase/OCI-style naming, like maintainer, version, or ideally `org.opencontainers.image.vendor`.
I am not sure about removing those since labels could break anyone who reads image metadata.

otel-integration/k8s-helm/tests/golden/ebpf-profiler.yaml
markshleifer-coralogix: The golden files are generated from helm template output for the high-risk preset values. The initial files were generated locally with Helm using the same chart, base values, preset values, release name, domain, and cluster name that the CI check uses.
I added an explicit update mode so this is now reproducible:
`.github/scripts/check-helm-golden-renders.sh --update`

It requires helm locally. The normal CI path still runs the same script without --update, which renders the manifests into a temp dir and diffs them against the checked-in golden files.

iblancasa: I see.
I'm asking because these files will need to be updated when a new collector version is released.

markshleifer-coralogix: Correct.

When the collector/chart version changes, these golden files are expected to change as well. The intended workflow is to run:
`.github/scripts/check-helm-golden-renders.sh --update`
and commit the resulting golden-file diffs together with the collector version bump.

.github/workflows/security-hygiene.yml
markshleifer-coralogix: As discussed, will do in a separate PR.

iblancasa (approved):

PR #969, opened 2026-07-16
feat(obi): add OBI to otel-ecs-ec2 and otel-linux-standalone

NimrodAvni78:
# Description

Adds **opt-in OpenTelemetry eBPF Instrumentation (OBI, v0.10.0)** to the two non-k8s shippers, coupled to each shipper's deploy mechanism so it isn't a separate product — mirroring the existing `otel-integration/k8s-helm` OBI subchart integration. (Requested in [this thread](https://coralogix-dev.slack.com/archives/C0BFVP2UXNU/p1784113346899789).)

### otel-ecs-ec2
- OBI runs as a **privileged sidecar** in the existing `coralogix_otel_agent` task (already host-net/host-pid/privileged with all eBPF host mounts), exporting to the node-local collector at `127.0.0.1:4317/4318`.
- Config authored in `values.yaml` (`obi.config`), rendered via `make obi-config` → `obi/obi-config.yaml`, written to the host by the launch-template user-data and bind-mounted read-only into the sidecar at `OTEL_EBPF_CONFIG_PATH`.
- Terraform vars: `enable_obi`, `obi_image`, `obi_image_version`, `obi_context_propagation`, `obi_config` (full override).

### otel-linux-standalone
- OBI installed as an **`obi.service` systemd unit (as root)**; the release binary is downloaded (arch-detected amd64/arm64) in `user_data.sh.tmpl`, mirroring the existing telemetrygen pattern.
- Same helm-rendered config flow (`make obi-config` → `obi/obi-config.yaml`), injected into user-data (base64+gzip).
- Terraform vars: `enable_obi`, `obi_version`, `obi_context_propagation`, `obi_config_path`.

### Shared
- **Full-parity OBI config with k8s** (ebpf buffer sizes, `payload_extraction` incl. GenAI, `redis_db_cache`, `prometheus_export`, `gen_ai` attribute select), dropping k8s-only pieces (Kubernetes metadata, k8s cache, k8s owner-name network filters) and using **host-process discovery** (`open_ports`/`exe_path`).
- **Disabled by default**; new `obi/` subfolder + README per target.
- Standalone chart versions bumped together (linux/windows/macos → `0.0.46`) to satisfy the standalone version-check; ecs chart → `0.0.41`.

# How Has This Been Tested?

Static validation (all passing):
- `helm lint` + `helm template` (default and `--set obi.enabled=true`) for both charts.
- `make obi-config` renders a valid OBI config for both targets; default `make otel-config` output is unaffected (OBI ConfigMap is gated behind `obi.enabled`).
- `tofu validate` for both terraform modules.
- The linux `user_data` template was rendered with `enable_obi=true` and passes `bash -n`.
- OBI config content is copied verbatim from the k8s v0.10.0 rendered config (same OBI version), with only the verified discovery-schema swap.

**Live smoke test (deploy each target with `enable_obi=true` and confirm OBI telemetry reaches Coralogix) is pending** and will be run before marking ready for review.

# Checklist:

🤖 Generated with [Claude Code](https://claude.com/claude-code)


NimrodAvni78:
## ✅ Live smoke test passed (both targets, Coralogix staging / `anemia-new`)

Deployed each target to AWS with `enable_obi=true`, confirmed OBI telemetry reaches Coralogix, then tore everything down (0 resources left).

### otel-linux-standalone
- `obi.service` installed via the release binary + systemd unit (as root), came up **active**.
- OBI loaded eBPF probes and discovered/instrumented host processes (incl. a test HTTP server, `type=python`).
- Collector exporter counters: `otelcol_exporter_sent_spans_total{exporter="coralogix"}` = **109**, `send_failed_spans_total` = **0**.

### otel-ecs-ec2
- OBI ran as the privileged sidecar in the collector task; discovered/instrumented all host processes — telemetry stamped `source="obi"`, `telemetry_distro_name="opentelemetry-ebpf-instrumentation"` (incl. `cdot` collector, ECS `agent`, `telemetrygen`, `containerd`, `dockerd`, `sshd`, …).
- Collector exported **24,914 spans** to staging with **0 failures**.

### Notes / findings
1. **(pre-existing, not OBI)** On `otel-linux-standalone`, the chart's full rendered collector config is incompatible with the pinned upstream `otelcol-contrib` v0.137.0 `.deb`: it uses the Coralogix-custom `systemd` receiver (absent upstream) plus `coralogix.domain_settings.keepalive.enabled` and `spanmetrics/db.add_resource_attributes` (rejected by that collector version). The base `make deploy` (without OBI) hits the same crash. The OBI smoke test used a minimal collector config to work around it. Worth a separate look at the deb/chart version alignment.
2. **(minor, ECS)** OBI logs a non-fatal warning `creating OTEL namespace in bpffs failed … mkdir /sys/fs/bpf/otel: no such file or directory`, disabling pinned-map extras (log enricher, profile correlation). Core instrumentation is unaffected. Possible follow-up: adjust the `bpf` host mount / pre-create the dir.
3. **CI:** the `Check Documentation formatting and links` failure is pre-existing external-link flakiness (`opentelemetry.io`/`kubernetes.io` timeouts) in files this PR does not touch; no errors in the added docs.

otel-linux-standalone/templates/obi-configmap.yaml
```diff
@@ -0,0 +1,19 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-obi
+  labels:
+    app.kubernetes.io/name: opentelemetry-ebpf-instrumentation
+    app.kubernetes.io/part-of: obi
+    app.kubernetes.io/component: config
+data:
+  obi-config.yml: |
```
povilasv: Why is this needed? Lets not add templates here they should be in helm-charts

otel-ecs-ec2/templates/obi-configmap.yaml
```diff
@@ -0,0 +1,19 @@
+{{- /*
+Renders the OpenTelemetry eBPF Instrumentation (OBI) config as a ConfigMap so
+that `make obi-config` can extract it (via `--set obi.enabled=true`) into
+obi/obi-config.yaml. It is gated behind .Values.obi.enabled so the default
+`make otel-config` / `make manifest` renders are unaffected.
+*/ -}}
+{{- if .Values.obi.enabled }}
+apiVersion: v1
```
povilasv: this should be in opentelemetry-helm-charts or in your subchart?

otel-linux-standalone/templates/obi-configmap.yaml
NimrodAvni78: im not sure how we usually template out this in linux and ecs?
should we use helm even if its not k8s? or terraform / something else

otel-ecs-ec2/templates/obi-configmap.yaml
NimrodAvni78: ill check but this just renders the config map even though its not k8s.
how should these configs be templated?

PR #987, opened 2026-08-09
test(otel-integration): cover the OTel eBPF profiler preset and validate its config against the pinned image

royfur:
Closes the test gaps that let the profiler crash ship. The crash itself is fixed upstream in [coralogix/opentelemetry-helm-charts#497](https://github.com/coralogix/opentelemetry-helm-charts/pull/497), which landed here via #988 (opentelemetry-collector 0.136.4, chart 0.0.335) — so this PR is now test coverage only, on top of that.

## Background

Profiling was broken in otel-integration 0.0.333: the chart rendered `tracers: all` on the `profiling` receiver, a key removed from the profiler in v0.156.0, so every profiler pod crash-looped with `'config.Config' has invalid keys: tracers`. 0.0.333 (#984) is where the image moved to 0.156.0.

## Two gaps let it through

1. **The broken path had no golden coverage.** The `ebpf-profiler` case uses `values-ebpf-profiler.yaml`, which enables the legacy `coralogix-ebpf-profiler` sub-chart — not the `opentelemetry-ebpf-profiler` sub-chart the wizard emits and that users actually run.
2. **Golden diffs structurally cannot catch this.** When upstream removes a config key, the rendered YAML stays byte-identical; only the binary's acceptance of it changes. The check was green while every profiler pod was crash-looping.

## Changes

- **New golden case `otel-ebpf-profiler`** built from `values-otel-ebpf-profiler.yaml`, i.e. the wizard configuration from the report (otlpExporter to the node agent, resourceDetection, fleetManagement without supervisor, `samplesPerSecond: 20`). The legacy `ebpf-profiler` case is left in place; it can be dropped when that sub-chart is retired.
- **New `.github/scripts/check-collector-config-validity.sh`**, wired into the `helm-golden-render` job. For every golden render it extracts the collector config from each ConfigMap and, for configs that declare the `profiling` receiver, runs `otelcol validate` against them. Details:
  - It runs inside **the profiler image the chart actually renders** (`ghcr.io/…/opentelemetry-collector-ebpf-profiler:0.156.0`, read out of the golden), so the check follows every dependency bump with no extra maintenance and needs no separate binary download.
  - `validate` unmarshals the config and builds the pipelines — no eBPF probes, no privileges, no cluster. It is the same code path that produced the crash message in production.
  - `${env:…}` references are stubbed, since `validate` resolves them. The config is passed in via `--config=env:…` rather than a bind mount, so it behaves the same in CI and on a laptop whose Docker VM doesn't share `$TMPDIR`.
  - Because pipelines are built for real, the `eks` resource detector constructs a Kubernetes client; a stub service-account token is written inside the container so it can. Nothing reads it and no request is made.
  - The script **fails if no `profiling` config is found** in the goldens, so this coverage cannot silently disappear again.

## Validation

Both checks run clean locally, and the new one was verified in both directions:

- `check-helm-golden-renders.sh` — clean, including the new case.
- `check-collector-config-validity.sh` — `OK otel-ebpf-profiler/…configmap_agent_yaml`.
- Re-inserting `tracers: all` into the golden makes it fail with exactly the production error: `'config.Config' has invalid keys: tracers`.
- `shellcheck` (v0.10.0, as CI pins) clean on both scripts; `helmlog validate` and `make check-docs` clean.

### DIFF
--- a/.github/scripts/check-collector-config-validity.sh
+++ b/.github/scripts/check-collector-config-validity.sh
@@ -1,0 +1,148 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Validates the collector configs embedded in the golden renders against the
+# container image the chart actually renders, using `otelcol ... validate`
+# (config unmarshal + pipeline construction — no eBPF probes, no cluster).
+#
+# Golden renders alone cannot catch an upstream config-key removal: the rendered
+# YAML stays byte-identical while the binary stops accepting it. That is how
+# `tracers: all` shipped in otel-integration 0.0.333 and crash-looped every
+# profiler pod with:
+#   'config.Config' has invalid keys: tracers
+#
+# Only configs that wire the `profiling` receiver are checked. The agent and
+# cluster-collector configs reference Coralogix-distribution-only components,
+# which no published upstream image can resolve.
+
+CHART_DIR="${CHART_DIR:-otel-integration/k8s-helm}"
+GOLDEN_DIR="${GOLDEN_DIR:-${CHART_DIR}/tests/golden}"
+
+require_cmd() {
+  command -v "$1" >/dev/null 2>&1 || { echo "Missing required command: $1" >&2; exit 1; }
+}
+require_cmd docker
+require_cmd awk
+
+tmpdir="$(mktemp -d)"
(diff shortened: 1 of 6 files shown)


povilasv:
@codex review

PR #1022, opened 2026-09-15
CX-58697: gate autoinstrumentation webhook E2E and add apache/nginx

cazorla19:
# Description

Adds Apache HTTPD and nginx cases to the existing no-CRD webhook E2E, and runs that job only when autoinstrumentation image pins (or the webhook harness) change — not on every Helm PR. Auto-bump PRs from #1020 will still exercise the languages they actually bump.

Fixes #CX-58697

# How Has This Been Tested?

- `bash scripts/autoinstrumentation-e2e-filter_test.sh` (pin-only, apache→nginx pairing, harness, README no-op, `--all`)
- `go test -c` for the e2e package
- Filter/harness change on this PR is expected to run the full webhook E2E in CI

# Checklist:
- [ ] I have updated the relevant Helm chart(s) version(s)
- [ ] I have updated the relevant component changelog(s)
- [x] This change does not affect any particular component (e.g. it's CI or docs change)

Made with [Cursor](https://cursor.com)

PR #1023, opened 2026-09-17
CX-58697: webhook E2E review follow-up

cazorla19:
# Description

Follow-up to #1022 review comments from [iblancasa](https://github.com/coralogix/telemetry-shippers/pull/1022#pullrequestreview-5211820392) (and the operator-chart gating note Douglas asked about).

- Drop the E2E exporter override to `http://coralogix-opentelemetry.default.svc:4317`. Operator 0.158.0 already forwards Spec env into the Apache/nginx attach init containers ([open-telemetry/opentelemetry-operator#5334](https://github.com/open-telemetry/opentelemetry-operator/pull/5334)), so the chart default `http://$(OTEL_NODE_IP):4317` is what this test should exercise.
- Keep both `apacheHttpd` and `apachehttpd` image pins. `apacheHttpd` is the CRD json tag and is ignored by `gopkg.in/yaml.v2` when the operator loads `--config-file`; without `apachehttpd` the attach init container is created with an empty image and kubelet fails the pod as `FailedCreate`. That belongs in the operator (`yaml:"apacheHttpd"` on `InstrumentationSpec`), not a further chart workaround. The 0.0.349 changelog for this pin already shipped in #1022, so this PR does not bump the chart again.
- Run the webhook E2E when the `opentelemetry-operator` subchart version in `Chart.yaml` changes (not on collector bumps).

Fixes #CX-58697

# How Has This Been Tested?

- `bash scripts/autoinstrumentation-e2e-filter_test.sh` (including operator chart bump → run all, collector bump → skip)

# Checklist:

Made with [Cursor](https://cursor.com)


otel-integration/k8s-helm/values.yaml
```diff
@@ -94,12 +94,17 @@ opentelemetry-autoinstrumentation:
                 value: http://$(OTEL_NODE_IP):4317
               - name: OTEL_EXPORTER_OTLP_PROTOCOL
                 value: grpc
+          # The Instrumentation CRD json tag is apacheHttpd. The no-CRD webhook
+          # loads this block with gopkg.in/yaml.v2, which ignores json tags and
+          # maps the Go field ApacheHttpd to apachehttpd. With only apacheHttpd,
+          # the attach init container is created with an empty image and kubelet
+          # fails the pod as FailedCreate (spec.containers[x].image: Required value).
+          # Keep both spellings until the operator adds yaml:"apacheHttpd" on the
+          # InstrumentationSpec field (that is an upstream fix, not a chart one).
```
iblancasa, line 103: I guess this should not be here.
We need to fix upstream

otel-integration/k8s-helm/e2e-test/testdata/values-e2e-instrumentation-webhook.yaml
```diff
@@ -1,14 +1,5 @@
 opentelemetry-autoinstrumentation:
   enabled: true
-  manager:
```
iblancasa, line 3: But Apache and Nginx still break on those defaults:
1. Spec env sets `OTEL_NODE_IP` with `valueFrom`
2. `yaml.v2` drops `valueFrom` when the operator loads the config file, the attach init gets `OTEL_NODE_IP=""`
3. Agent conf becomes `http://:4317`.
 
Please also remove the `OTEL_NODE_IP` `valueFrom` entry from `values.yaml` (keep sampler/logs) and rely on the operator default. I verified apache/nginx pass after that.

We need to fix upstream. I will do

